PR #4, opened 2020-06-02
Apply theming & add bottom navigation bar

marfavi:
Applied our colours to the Scaffold/MaterialApp.
Added pages & a way to navigate through them. (Only the Tickets page have content, but they all have their own title, seen in the app bar).
image

### DIFF
--- a/lib/base/strings.dart
+++ b/lib/base/strings.dart
@@ -1,0 +1,8 @@
+abstract class Strings {
+  static const appTitle = 'Cafe Analog';
+
+  static const ticketsPageTitle = 'Tickets';
+  static const receiptsPageTitle = 'Receipts';
+  static const statsPageTitle = 'Statistics';
+  static const settingsPageTitle = 'Settings';
+}
--- a/lib/base/strings.dart
+++ b/lib/base/strings.dart
@@ -12,4 +12,7 @@ abstract class Strings {
   static const receiptsNavTitle = receiptsPageTitle;
   static const statsNavTitle = 'Stats';
   static const settingsNavTitle = settingsPageTitle;
+
+  // Test strings.
+  static const ticketsPageTestString = 'Test string for Tickets page ☕';
 }
--- a/lib/base/strings.dart
+++ b/lib/base/strings.dart
@@ -1,8 +1,15 @@
 abstract class Strings {
   static const appTitle = 'Cafe Analog';
 
+  // Titles for the app bar.
   static const ticketsPageTitle = 'Tickets';
   static const receiptsPageTitle = 'Receipts';
   static const statsPageTitle = 'Statistics';
   static const settingsPageTitle = 'Settings';
+
+  // Titles for the navigation bar.
+  static const ticketsNavTitle = ticketsPageTitle;
+  static const receiptsNavTitle = receiptsPageTitle;
+  static const statsNavTitle = 'Stats';
+  static const settingsNavTitle = settingsPageTitle;
 }
--- a/lib/base/style/colors.dart
+++ b/lib/base/style/colors.dart
@@ -15,7 +15,7 @@ abstract class AppColor {
   static const Color error = Color(0xffFF4D00);
 
   /// Creates a swatch from a given Color.
-  /// https://medium.com/@filipvk/creating-a-custom-color-swatch-in-flutter-554bcdcb27f3
+  //  https://medium.com/@filipvk/creating-a-custom-color-swatch-in-flutter-554bcdcb27f3
   static MaterialColor createMaterialColor(Color color) {
     List strengths = <double>[.05];
     Map swatch = <int, Color>{};
--- a/lib/base/style/colors.dart
+++ b/lib/base/style/colors.dart
@@ -9,8 +9,30 @@ abstract class AppColor {
   static const Color white = Color(0xffFAFAFA);
   static const Color gray = Color(0xff767676);
   static const Color lightGray = Color(0xffC2C2C2);
-  
+
   static const Color slightlyHighlighted = Color(0xffFEF4D2);
   static const Color success = Color(0xff4CAF50);
   static const Color error = Color(0xffFF4D00);
+
+  /// Creates a swatch from a given Color.
+  /// https://medium.com/@filipvk/creating-a-custom-color-swatch-in-flutter-554bcdcb27f3
+  static MaterialColor createMaterialColor(Color color) {
+    List strengths = <double>[.05];
+    Map swatch = <int, Color>{};
+    final int r = color.red, g = color.green, b = color.blue;
+
+    for (int i = 1; i < 10; i++) {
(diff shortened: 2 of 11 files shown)


lib/main.dart
```diff
@@ -1,119 +1,98 @@
...
   runApp(MyApp());
 }
 
 class MyApp extends StatelessWidget {
-  // This widget is the root of your application.
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
-      title: 'Flutter Demo',
+      title: 'Café Analog',
```
jonasanker, line 19: We're moving over to call us `Cafe Analog` without the `é`

lib/main.dart
```diff
@@ -1,119 +1,98 @@
...
-        primarySwatch: Colors.blue,
-        // This makes the visual density adapt to the platform that you run
-        // the app on. For desktop platforms, the controls will be smaller and
-        // closer together (more dense) than on mobile platforms.
+        primarySwatch: AppColor.createMaterialColor(AppColor.primary),
+        primaryColor: AppColor.primary,
+        brightness: Brightness.dark,
+        backgroundColor: AppColor.background,
         visualDensity: VisualDensity.adaptivePlatformDensity,
       ),
```
jonasanker, line 26: I suggest this constructor is moved to a static method in a new class in the styles package, to separate concerns and not clutter up the Main class.

test/widget_test.dart
```diff
@@ -15,16 +15,16 @@ void main() {
     // Build our app and trigger a frame.
```
jonasanker, line 15: The test description no longer matches what actually is being tested, so that should be changed.

```suggestion
'Sample test'
```
However as another note, this test only serves as a dummy test as it should be expanded should it actually test something properly. I suggest we'll just let it be for the time being a come by and clean up at a later point.

lib/widgets/pages/tickets_page.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:coffeecard/base/style/text_styles.dart';
+import 'package:flutter/material.dart';
+
+class TicketsPage extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Center(
+      child: Text('Buy and spend tickets here!', style: AppTextStyle.sectionTitle),
```
jonasanker, line 8: I know this is properly just a test string but we can just as well start up some good practices. According to [flutter documentation](https://flutter.dev/docs/get-started/flutter-for/android-devs#where-do-i-store-strings-how-do-i-handle-localization) as `static final` or `const` fields in its own class.

I suggest we create a new class `StringsConstants` or something similar under a new subpackage in the `base` package.

lib/base/style/colors.dart
```diff
@@ -9,8 +9,30 @@ abstract class AppColor {
...
   static const Color gray = Color(0xff767676);
   static const Color lightGray = Color(0xffC2C2C2);
-  
+
   static const Color slightlyHighlighted = Color(0xffFEF4D2);
   static const Color success = Color(0xff4CAF50);
   static const Color error = Color(0xffFF4D00);
+
+  /// Creates a swatch from a given Color.
+  /// https://medium.com/@filipvk/creating-a-custom-color-swatch-in-flutter-554bcdcb27f3
```
jonasanker, line 18: It's cool with `DOC comments`! Keep them coming
However I think the link medium should just be a regular comment as it gives no value in autogenerated code documentation.

lib/main.dart
jonasanker: This should also be moved to the StringConstants as suggested in the comment for `lib/widgets/pages/tickets_page.dart`

lib/main.dart
```diff
@@ -1,119 +1,98 @@
...
-
+class HomePage extends StatefulWidget {
   @override
-  _MyHomePageState createState() => _MyHomePageState();
+  _HomePageState createState() => _HomePageState();
 }
 
-class _MyHomePageState extends State<MyHomePage> {
-  int _counter = 0;
+class _HomePageState extends State<HomePage> {
```
jonasanker, line 37: I think the HomePage should be extracted to its own class instead of a nested class.

jonasanker:
Also please note that your branch does not have the latest changes from master. You might want to rebase from master before continuing.. You would that in the terminal like this:

```bash
# your feature branch should be checked out
git fetch
git rebase origin/master

# resolve any conflicts and voila - it's rebased !
```

test/widget_test.dart
marfavi: Changed description & added more comments for the test.

lib/widgets/pages/tickets_page.dart
marfavi: Added Strings in lib/base.

lib/main.dart
marfavi: ThemeData moved to lib/base/style/theme.dart as analogTheme.

jonasanker:
I can see that you have added a duplicate commit 866cb6ecf31afc2f42d43e2ce2c6a29000ed94a4 which is the same as the one in master. I can see what you tried achieving but it is much better to do a rebase from master than creating new commits for the same change :)

You can do a rebase from my like this:
```bash
# You feature branch must be checked out
git fetch
git rebase origin/master
```
A conflict might arise since you have two commits making the same change. In that case dropping the 866cb6ecf31afc2f42d43e2ce2c6a29000ed94a4 commit would be better. You can do this in an interactive rebase - I can show you how

lib/base/strings.dart
```diff
@@ -0,0 +1,18 @@
...
+
+  // Titles for the app bar.
+  static const ticketsPageTitle = 'Tickets';
+  static const receiptsPageTitle = 'Receipts';
+  static const statsPageTitle = 'Statistics';
+  static const settingsPageTitle = 'Settings';
+
+  // Titles for the navigation bar.
+  static const ticketsNavTitle = ticketsPageTitle;
+  static const receiptsNavTitle = receiptsPageTitle;
```
jonasanker, line 12: Minor comment - you don't have to change if it's by a good reason.. I'm not sure why you don't just use the page title where needed instead of creating a new constant?

lib/main.dart
```diff
@@ -1,119 +1,21 @@
...
-        ),
-      ),
-      floatingActionButton: FloatingActionButton(
-        onPressed: _incrementCounter,
-        tooltip: 'Increment',
-        child: Icon(Icons.add),
-      ), // This trailing comma makes auto-formatting nicer for build methods.
+      title: Strings.appTitle,
+      theme: analogTheme,
+      home: HomePage(),
```
jonasanker, line 18: Lovely clean. Excellent :muscle:

lib/base/strings.dart
marfavi: It's mostly for flexibility. If we were to implement another language, we may use shortened names for other pages 🙂

PR #6, opened 2020-06-27
RestClient, Authentication and sample AppConfig Repository

jonasanker:
- Sample RestClient using Dio, Retrofit and JsonSerializable packages
- Client and models are implemented as abstract classes and the implementation is generated using build generators
- Updated CI to generate code to make the build pass
- Run build command to generate source code `flutter pub run build_runner build`
- SecureStorage to save/read token
- AuthenticationInterceptor adds Authorization Header if token is stored


lib/persistence/http/RestClient.dart
```diff
@@ -0,0 +1,21 @@
...
+import 'package:coffeecard/model/Token.dart';
+import 'package:dio/dio.dart';
+import 'package:retrofit/retrofit.dart';
+
+part 'RestClient.g.dart';
+
+@RestApi()
+abstract class RestClient {
+  static const String PRODUCTION_URL = "https://analogio.dk/clippy";
+  static const String DEBUG_URL = "https://beta.analogio.dk/api/clippy";
```
TTA777, line 12: Since we normally refer to this as the test environment, is there any reason why we want to call it debug all of a sudden?

lib/persistence/repositories/impl/_AccountRepository.dart
```diff
@@ -0,0 +1,31 @@
...
+  final SecureStorage _storage;
+
+  _AccountRepository(this._restClient, this._logger, this._storage);
+
+  @override
+  Future login(String userName, String password) async {
+    var login = Login(userName, password, "");
+
+    var token = await _restClient.login(login).catchError((Object obj) {
+      switch (obj.runtimeType) {
```
TTA777, line 20: Any reason why we are not using a standard try - catch for error handling?

lib/persistence/http/RestClient.dart
```diff
@@ -0,0 +1,21 @@
...
+
+  factory RestClient(Dio dio, {String baseUrl}) = _RestClient;
+
+  @GET("​/api​/v1​/AppConfig")
```
TTA777, line 16: I personally don't like that we will end up having every single endpoint in the rest-client. Is this really the recommended way to do it?

lib/persistence/repositories/impl/_AccountRepository.dart
```diff
@@ -0,0 +1,31 @@
...
+import 'package:coffeecard/persistence/storage/SecureStorage.dart';
+import 'package:dio/dio.dart';
+import 'package:logger/logger.dart';
+
+class _AccountRepository implements AccountRepository {
+  final RestClient _restClient;
+  final Logger _logger;
+  final SecureStorage _storage;
+
+  _AccountRepository(this._restClient, this._logger, this._storage);
```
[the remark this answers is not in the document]
jonasanker: I withdraw the `_` naming in a later commit but that's on the dependency injection and test branch.

lib/persistence/repositories/AccountRepository.dart
```diff
@@ -0,0 +1,3 @@
+abstract class AccountRepository {
```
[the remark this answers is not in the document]
jonasanker: There's no naming convention for interfaces in Dart. Java has it like this, C# has it with an `I` as prefix. In Dart everything is an interface so that's even more funky.

Regardless, I have named the implementation with `Impl` (like Java convention) to distinct the implementation from the interface. I would suggest that's the way to go but we would agree on a naming convention 👌

lib/persistence/repositories/impl/_AccountRepository.dart
TTA777: I honestly don't see where on that page it says this is the recommended way to do it. It even has a pretty large disclaimer saying:

"You don’t need this page if your code uses the language’s asynchrony support — async, await, and error handling using try-catch. For more information, see the asynchronous programming codelab."

Is it because retrofit is "old fashioned" and reliant on using future methods?

lib/persistence/http/RestClient.dart
TTA777: That's unfortunate :/ How much boilerplate code are we avoiding by using retrofit instead of using dio directly, where we have a bit more freedom?

lib/persistence/http/RestClient.dart
TTA777: That probably be ideal in the long run if we do that instead.

lib/persistence/repositories/AccountRepository.dart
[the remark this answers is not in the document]
TTA777: Personally I would favour the c# way of doing it if there is no convention in dart, simply because our backend is in c# so it might be more recognisable for anyone working on both systems. Either way works of course, as long as we settle on a convention and stick with it 😃

lib/persistence/http/RestClient.dart
jonasanker: Retrofit generates a lot of boilerplate for us (it can easily be some hundred lines) in the long term. I think it's quite neat. I'm not sure what we are missing and as we can make custom Dio factories we are able to override certain behaviour if needed.

jonasanker: Try to run the script locally and take a look in what is generated just to give you an idea :-)

lib/persistence/repositories/impl/_AccountRepository.dart
jonasanker: Good found with the disclaimer. I guess it's a taste which way you like. In this case you're right that the try/catch block makes more sense. I'll rewrite

lib/persistence/repositories/AccountRepository.dart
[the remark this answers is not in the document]
jonasanker: If everything is an interface in Dart, then one can consider if we even need abstract classes for our Repositories. 🤔

lib/persistence/http/RestClient.dart
TTA777: Only thing I was missing was not having the methods in the rest client, but I suppose I can live with it :/

lib/persistence/repositories/AccountRepository.dart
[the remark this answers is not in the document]
TTA777: We probably don't strictly need it, but it might be nice to have or is there a reason you are considering removing them?

lib/persistence/repositories/impl/_AccountRepository.dart
TTA777: Good to hear, I take it you will be rewriting the appconfig one as well then?

lib/persistence/http/RestClient.dart
jonasanker: Haha :sweat_smile: Well, thats the Retrofit core, so that's how it is

jonasanker: I have renamed the RestClient to `CoffeeCardApiClient` to emphasize that it is not a generic client but a specific one for the CoffeeCard API.

lib/persistence/repositories/impl/_AccountRepository.dart
jonasanker: Done. See latest commit

lib/persistence/http/RestClient.dart
jonasanker: I'll create it as a Trello task so we can configure out how to do it properly

lib/persistence/repositories/impl/account_repository_impl.dart
```diff
@@ -0,0 +1,28 @@
...
+  Future login(String userName, String password) async {
+    var login = Login(userName, password, "");
+
+    try {
+      var token = await _restClient.login(login);
+      await _storage.saveToken(token.token);
+    } on DioError catch (error) {
+      final httpResponse = error.response;
+      _logger.e("API Error ${httpResponse.statusCode} ${httpResponse.statusMessage}");
+      rethrow;
```
jonasanker, line 25: A proper way of handling rest errors will be addressed in https://trello.com/c/ZV5Ul2N4/33-error-handling-in-repositories

lib/persistence/http/RestClient.dart
jonasanker: Trello task https://trello.com/c/qszXRpMk/40-set-restclient-url-in-build-configuration

PR #16, opened 2021-12-06
Rewrite authentication/login flow

### LINKED ISSUE #12: Do not use onPage in LoginState
Refactor `LoginState` to not be responsible for handling the `onPage`. This should be the responsibility of the UI. Possibly introduce inheritance to allow reuse of the login page widgets.

### LINKED ISSUE #15: Reduce loading time on LoginScreen
After successful login, the app makes a call to retrieve the user information. Change behaviour

**Tasks**
1. Move navigation from `authentication_bloc`/`authentication_service` to `login_bloc`
2. Retrieve user information REST call should be moved to the home page

marfavi:
# Rewrite authentication/login flow
Resolves #15  , resolves #12 

## In short
*  __Login flow now works correctly, in a less hacky way__
*  __A lot of opinionated changes & changes to the architecture__
* No longer uses `onPage` state to determine where the user is in the login flow
* Only loads the user token in the login flow. User data should be loaded afterwards.

## Code changes
* Blocs rewritten
  * Now use the new syntax (`on<Event>`) instead of the deprecated one (`mapEventToState`)
  * Event names changed & added more events
  * Bloc state classes no longer uses class inheritance
* Navigation is handled by new *Name*Router widgets (`auth_router`, `login_router`)
  * These look at their respective Blocs to decide what page to show on screen
* `AuthenticationService` was removed.
* `LoginBloc` now has access to `AuthBloc` and `AccountRepository`
  * Calls `AccountRepository.login()` to send a log in request
  * Adds an `AuthEvent` on successful login
* `AuthBloc` now has access to Secure Storage, which is accessed whenever the user gets authenticated/unauthenticated

Here's how the new Blocs interact with each other (some fields/events might be slightly changed since this drawing was made):

image

## Misc
* Changed `analysis_options.yaml` to be more strict
  * A lot of small changes have occurred in some files as a result
* App remembers if the user is logged in, and will skip the login flow if credentials are saved. Intended for demonstrative purposes only, but I see a possibility of discussing whether this behavior should stay in the app.

## Notable file changes
`lib/blocs/authentication/authentication_bloc.dart`
`lib/blocs/authentication/authentication_event.dart`
`lib/blocs/authentication/authentication_state.dart`

`lib/blocs/login/login_bloc.dart`
`lib/blocs/login/login_event.dart`
`lib/blocs/login/login_state.dart`

`lib/main.dart`

`lib/persistence/repositories/impl/account_repository_impl.dart` (deleted)
`lib/persistence/repositories/account_repository.dart` (this is now the implementation)

`lib/persistence/repositories/authentication_service.dart` (deleted)

`lib/persistence/storage/secure_storage.dart`

`lib/utils/exception_extractor.dart`

`lib/widgets/routers/auth_router.dart`
`lib/widgets/routers/entry_router.dart`
`lib/widgets/routers/login_router.dart`


jonasanker (approved):
This is fine 🔥

marfavi (commented):
Approve pls

Issue #62, opened 2022-01-12
Error message format when username + password mismatch


fremartini:
Format should be fixed

![image](https://user-images.githubusercontent.com/39860858/149093976-1de2f5f3-9b19-4152-9556-3211c4a1d066.png)

jonasanker:
I think in general we will need to have a small discussion on how the app should react to API errors

PR #71, opened 2022-01-22
Improve register flow

marfavi:
Most important changes of this PR are:

* Register flow completes fully, catches errors and navigates out of the register flow when registered
  * Format of error is incorrect (see #62)
* Nested routing has been introduced through `RegisterFlow`
  * replaces `RegisterRouter`, which relied on bloc state to show a corresponding screen
  * `RegisterFlow` contains a single Scaffold, one BlocProvider, and a Navigator with a public static `push` method to navigate between routes (these routes being enter email, enter passcode, enter name).
  * Each route in the register flow is now only a section title and text field, in contrast to an entire Scaffold with content.
  * Nice animations between each route in the register flow.
* Replaced `RegisterBloc` with `RegisterCubit`

All changes can be found in the commit message of 3a45c8345296b798789ca6a915dd428ff0dfd8d4


HeeysamH (approved):
Added correct encoding when registering a user.

lgtm now.

PR #90, opened 2022-02-04
bump versions and depend on concrete version

### LINKED ISSUE #77: Specify exact version numbers instead of ranges in pubspec.yaml
From https://github.com/AnalogIO/coffeecard_app/pull/65#discussion_r795042638

We should use specific versions in the `pubspec.yaml` instead of relative, like e.g. higher than version x.
Motivation: Having relative versions makes us vulnerable to possible breaking changes from future package updates

fremartini:
close #77

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -25,58 +25,58 @@ dependencies:
     sdk: flutter
 
   # ui
-  gap: ^2.0.0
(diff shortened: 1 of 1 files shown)


TTA777 (approved):
LGTM

Issue #94, opened 2022-02-05
Use Either for error handling in repositories and as a general pattern


jonasanker:
- Decide on general error types for api errors, network etc
- Implement Either in all repositories

PR #100, opened 2022-02-05
Feature/frem/either

fremartini:
closing #94

### DIFF
--- a/lib/utils/either.dart
+++ b/lib/utils/either.dart
@@ -1,0 +1,21 @@
+abstract class Either<L, R> {
+  const Either();
+
+  bool get success => this is Right<L, R>;
+  bool get failure => this is Left<L, R>;
+
+  L get left => (this as Left<L, R>)._l;
+  R get right => (this as Right<L, R>)._r;
+}
+
+class Left<L, R> extends Either<L, R> {
+  final L _l;
+
+  const Left(this._l);
+}
+
+class Right<L, R> extends Either<L, R> {
+  final R _r;
+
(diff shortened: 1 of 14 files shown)


lib/widgets/pages/home_page.dart
```diff
@@ -34,7 +34,12 @@ class _HomePageState extends State<HomePage> {
...
     try {
-      config = await _configRepository.getAppConfig();
+      final either = await _configRepository.getAppConfig();
+      if (either.success) {
+        config = either.right;
+      } else {
+        config = null;
+      }
     } catch (e) {
       config = null;
```
jonasanker, line 44: Do we still need the try catch? Can we avoid to have a null config?

lib/utils/either.dart
```diff
@@ -0,0 +1,21 @@
+abstract class Either<L, R> {
+  const Either();
+
+  bool get success => this is Right<L, R>;
+  bool get failure => this is Left<L, R>;
```
jonasanker, line 5: To follow the usual implementation pattern of Either, I suggest these should be called `isLeft` and `isRight`

jonasanker (changes requested):
Minor comments on `Either`

lib/widgets/pages/home_page.dart
fremartini: This code snippet will be removed in another PR which is why i did not bother to fix it :)

jonasanker (approved):
Approved ✅

PR #87, opened 2022-02-02
Feature/frem/move test db widget

fremartini:
close #86, close #85, close #62

PR #104, opened 2022-02-07
Rewrite EnvironmentCubit; Remove TestDBWidget (Changes to #87)

marfavi:
`EnvironmentCubit` is placed at the top of the widget tree, so all widgets can know what environment we are connected to. Furthermore, no widgets are shown until the environment is known.

Test environment warnings should be handled on a per-widget basis. Examples:
* The login page could have a "Logging in to test environment" warning at the top of the screen
* The logo on the login page could be changed based on environment (Logo variation with a "Beta/Test" sign hanging off it).
* Warning dialogs shown before each ticket swipe or purchase.

Removed the Test DB widget as a result.

Discuss

## Commit changes

`EnvironmentCubit`
  - Rename from `EnvironmentTypeCubit` to  `EnvironmentCubit`
  - Remove UI related method and field
  - Replace state class with `Environment` enum (w/ extension)
  - Move `_configRepository` to constructor signature
  - Add FIXME to handle API error for `getConfig()`

`AuthenticationStatus`
  - Rename `AuthStatus` to `AuthenticationStatus`
  - Move `AuthenticationStatus` to `authentication_state.dart`
  - Add helper extension to `AuthenticationStatus`

Add `SplashRouter`
  - Will show splash page until both env and auth status is known
  - replaces `AuthRouter`

Remove `AuthRouter`
Remove `DatabaseConnectionWidget`
Adjust `main.dart` and `login_page.dart` to reflect these changes

Remove logo from `SplashPage`


jonasanker:
@fremartini I agree on @fremartini point here. I would also like the status was visible all the time. 

I'm definitely open for finding another way of showing it. I think the "new way" of showing it should be part of this PR

jonasanker (changes requested):
See comment

marfavi:
@fremartini 
> (...) I do however not agree with only displaying a message at the login screen and when a purchase is about to happen (...)

The examples I gave were just that, examples. The list is non-exhaustive and any implementation would let the tester know they were connected to the test environment.

@jonasanker 
Any solution, including overlaying a portion of the screen in an elegant way, will require work that is out of scope of this branch.

Furthermore, we need #87 merged to develop ASAP, as it contains changes in 61 files (not including this branch) and is blocking any further progress in other branches.

PR #109, opened 2022-02-08
Add Settings flow and related widgets

marfavi:
Contains the logout button at the correct place, but it is the only button that works in this flow.

The rest is WIP


fremartini:
looks like the build is failing because of an unused variable

.gitignore
```diff
@@ -52,3 +52,5 @@ app.*.map.json
 
 # test coverage file
 /coverage/lcov.info
+ios/Podfile
+ios/Podfile.lock
```
jonasanker, line 56: To my knowledge, these should not be gitignored as they contain package version information for iOS or am I wrong?

marfavi: You are right, we ignored these files out of caution before researching their importance.

I'm removing the .gitignore entries here, but I'd like to see the podfiles added in a seperate branch

marfavi: See #114

jonasanker: Perfect, all good 👍🏻

lib/widgets/pages/settings/settings_page.dart
```diff
@@ -0,0 +1,101 @@
...
+          } else {
+            return Column(
+              children: [
+                Padding(
+                  padding: const EdgeInsets.only(left: 16, right: 16, top: 16),
+                  child: UserCard(state),
+                ),
+                const Gap(16),
+                SettingsGroup(
+                  title: 'Account',
```
jonasanker, line 34: Move strings here and below to our strings class :)

lib/data/repositories/v1/account_repository.dart
```diff
@@ -75,6 +75,16 @@ class AccountRepository {
     }
   }
 
+  Future<void> updatePasscode(String passcode) async {
```
jonasanker, line 78: This should be an `Either` of e.g. `Either<ApiError, boolean>` maybe? You can probably just return the either from `UpdateUser()`

lib/models/account/user.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  const User({
+    required this.id,
+    required this.name,
+    required this.email,
+    required this.privacyActivated,
+    this.programmeId,
+  });
+
+  User.fromDTO(UserDto dto)
```
jonasanker, line 19: I think it would be more suitable if the Account Repository returned the user object you needed instead of mapping it to another object outside the repository.

lib/cubits/settings/settings_cubit.dart
```diff
@@ -0,0 +1,21 @@
...
+
+part 'settings_state.dart';
+
+class SettingsCubit extends Cubit<SettingsState> {
+  final AccountRepository _accountRepository;
+
+  SettingsCubit(this._accountRepository) : super(const SettingsState());
+
+  Future<void> loadUser() async {
+    final user = User.fromDTO(await _accountRepository.getUser());
```
jonasanker, line 14: Let the account repository take care of this mapping unless we have a reason to use the dto somewhere else :)

lib/widgets/pages/settings/your_profile_page.dart
```diff
@@ -0,0 +1,72 @@
...
+        Text(
+          user.name,
+          style: AppTextStyle.sectionTitle,
+          textAlign: TextAlign.center,
+        ),
+        const Gap(8),
+        Text('BSWU student', style: AppTextStyle.explainer),
+        const Gap(24),
+        SettingsGroup(
+          title: 'Edit profile',
```
jonasanker, line 50: Move strings to strings class

jonasanker (changes requested):
A few quick changes suggested

lib/data/repositories/v1/account_repository.dart
marfavi: Now returns `Future<Either<ApiError, void>>`

lib/cubits/settings/settings_cubit.dart
marfavi: `getUser` now returns `Future<Either<ApiError, User>>` so this has been fixed

lib/models/account/user.dart
marfavi: `User.fromDTO` is now only called inside `AccountRepository`

marfavi:
Oh no, I had auto-merge on
:(

Issue #107, opened 2022-02-07
Warning: Mapping new ns (...) to old ns (...)


marfavi:
Anyone know what this means?

Console output when starting app:
```
Launching lib\main.dart on Android SDK built for x86 in debug mode...
lib\main.dart:1
Warning: Mapping new ns [http://schemas.android.com/repository/android/common/02]() to old ns [http://schemas.android.com/repository/android/common/01]()
Warning: Mapping new ns [http://schemas.android.com/repository/android/generic/02]() to old ns [http://schemas.android.com/repository/android/generic/01]()
Warning: Mapping new ns [http://schemas.android.com/sdk/android/repo/addon2/02]() to old ns [http://schemas.android.com/sdk/android/repo/addon2/01]()
Warning: Mapping new ns [http://schemas.android.com/sdk/android/repo/repository2/02]() to old ns [http://schemas.android.com/sdk/android/repo/repository2/01]()
Warning: Mapping new ns [http://schemas.android.com/sdk/android/repo/sys-img2/02]() to old ns [http://schemas.android.com/sdk/android/repo/sys-img2/01]()
```

jonasanker:
This could look like a Gradle issue according to https://github.com/flutter/flutter/issues/87464 and [Warning: Mapping new ns to old ns and emulator stopping abruptly](https://stackoverflow.com/questions/68789848/warning-mapping-new-ns-to-old-ns-and-emulator-stopping-abruptly).

Solution should be up update an align Gradle versions in `gradle-wrapper.properties` and `/android/build.gradle`

PR #110, opened 2022-02-08
bump gradle version to 7.3

fremartini:
closing #107 

as correctly stated by @jonasanker updating the gradle version fixed the warnings related to 'mapping new ns (...) to old ns (...)'

### DIFF
--- a/android/gradle/wrapper/gradle-wrapper.properties
+++ b/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,5 @@
-#Fri Jun 23 08:50:38 CEST 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-all.zip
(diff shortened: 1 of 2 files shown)


marfavi:
Works on my machine 👍🏽

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Fri Jun 23 08:50:38 CEST 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-bin.zip
```
jonasanker, line 3: What's the difference between the all and bin version of gradle?

jonasanker: Also. Can we align the versions, so they use the same version of 7.x?

jonasanker (commented):
Comments on versioning

android/gradle/wrapper/gradle-wrapper.properties
fremartini: @jonasanker it seems they are independent. Looking through the gradle settings in Android Studio it seems a new version of gradle was released yesterday (7.4) 

image

jonasanker: All good then. I'll approve

jonasanker (approved):

marfavi:
Squash these commits? :)

Issue #91, opened 2022-02-04
receiptspage throws exception upon loading


fremartini:
![image](https://user-images.githubusercontent.com/39860858/152565952-2c374006-b1cf-4481-b999-036cb5833b0f.png)

Upon loading the 'Receipts' from the tabs the program throws an exception (receipt_repository.dart l. 56)

fremartini:
caused by user being unauthorized, fix #68

PR #112, opened 2022-02-08
tickets_repository now returns empty lists if API call fails

fremartini:
closing #91


TTA777 (changes requested):
Seems the two to-dos no longer apply to this version of the code, no? I would remove them if that is the case. 

Since you are making changes anyway, you might as well also update the repository to make use of Either as the return type.

Also, not sure if we want to do more to actually check for null, since it will currently throw an exception if a null somehow occurs in the body/ values. Currently, iirc the API is set up to enforce no values would be null, so probably not an issue

lib/data/repositories/v1/receipt_repository.dart
```diff
@@ -10,20 +12,16 @@ class ReceiptRepository {
 
   /// Retrieves all of the users receipts
   /// This includes both their used tickets and purchased tickets
-  Future<List<Receipt>> getUserReceipts() async {
-    final usedTicketRequest = _api.apiV1TicketsGet(used: true);
-    final purchasesRequest = _api.apiV1PurchasesGet();
-
-    final usedTicketResponse = await usedTicketRequest;
-    final purchaseResponse = await purchasesRequest;
+  Future<Either<ApiError, List<Receipt>>> getUserReceipts() async {
```
TTA777, line 15: Any reason this was changed? The original intention was to send both requests out, and then wait for them. Reasoning being, it is the waiting time that will be the longest, so we might as well send all requests first and then wait, since any waiting done at that point should affect both requests. Not just the first one sent.

TTA777 (approved):
lgtm

PR #115, opened 2022-02-09
Move strings from settings pages to `strings.dart`

marfavi:
To complete PR #109 that was merged too early


jonasanker (approved):
LGTM 👍🏻

PR #122, opened 2022-02-11
Feature/frem/statistics

fremartini:


lib/widgets/components/cards/leaderboard_card.dart
```diff
@@ -0,0 +1,76 @@
...
+import 'package:gap/gap.dart';
+
+class LeaderboardCard extends CardBase {
+  //FIXME: implement in a smarter way once user data is available
+  final int rank;
+  final String name;
+  final String programme;
+  final int cups;
+
+  LeaderboardCard({
```
marfavi, line 15: This should be based on `ListEntry`, not `CardBase`.

lib/widgets/components/cards/card.dart
```diff
@@ -49,6 +50,9 @@ abstract class CardBase extends StatelessWidget {
   /// If not null, the card will splash when tapped.
   final VoidCallback? onTap;
 
+  // Radius of the card border, lower values are sharper.
+  final double borderRadius;
```
marfavi, line 54: All cards have the same border radius, so this should not be a field.

lib/widgets/components/cards/stat_card.dart
```diff
@@ -0,0 +1,28 @@
...
+  final String title;
+  final String value;
+
+  StatisticsCard(this.title, this.value)
+      : super(
+          color: AppColor.white,
+          top: Column(
+            children: [
+              Text(title, style: AppTextStyle.textField),
+              const Gap(10),
```
marfavi, line 18: Use the `gap: double` parameter that belongs to `CardBase` instead of using a `Column` and `Gap`.

marfavi (commented):
Some comments.

Furthermore, I disagree with having a `cards` folder.
Instead, we should group files by flow.
We already have
- receipt for Receipts flow
- entry/login for Login flow
- entry/register for Register flow
so I suggest to have all Statistics widgets in a `statistics` folder.

marfavi:
This branch contains too many changes. Separate your changes into multiple branches (and pull requests) instead.

We should strive to put general changes (changes that will impact multiple flows or aspects of the app) in separate branches.

  - Don't move files you aren't working on (in this case, only move statistics-related files--don't move cards from other flows)
  - Generalize the filter bar in a separate branch, make a PR, and then base the rest of your work off that branch
  - Your grid implementation should also sit in its own branch (although I have a more complete implementation based off yours in [feature/omid/grid](https://github.com/AnalogIO/coffeecard_app/tree/feature/omid/grid))

Issue #68, opened 2022-01-18
Check if current bearer token is expired


TTA777:
Currently, we don't check if the bearer token is expired, which results in the user appearing logged in, but the requests fail.

Two potential solutions could be to either:
- Check in the AuthBloc whether the retrieved token is valid, and has a reasonable time left to be used in
- Check in the authentication interceptor if the token is valid on requests

Another question is, what behaviour do we want when we see an expired token, i.e. do we want to either:
- Kick the user back to the login screen?
- Use stored user credentials to get a new bearer token from the API?

jonasanker:
The Chopper library we're using as Http client, supports an authentication challenge class. 

The documentation is a little sparse. My understanding is that in the case a request Http fails with a Http 401 request, the authenticator class can be implemented to e.g. retry a request, call an API to get a new token etc.

References
- [Authenticator class](https://pub.dev/documentation/chopper/latest/chopper/Authenticator-class.html)
- [Flutter Chopper 401 renew & retry interceptor](https://stackoverflow.com/questions/67021327/flutter-chopper-401-renew-retry-interceptor)

PR #138, opened 2022-02-13
Feature/frem/check bearer token

fremartini:
closing #68 

* Upon 401 from API app tries to get a new token with credentials from secure storage
* User is signed out if sufficient amount of retries has been attempted

### DIFF
--- a/lib/utils/reactivation_authenticator.dart
+++ b/lib/utils/reactivation_authenticator.dart
@@ -1,0 +1,38 @@
+import 'dart:async';
+
+import 'package:chopper/chopper.dart';
+import 'package:coffeecard/data/repositories/v1/account_repository.dart';
+import 'package:coffeecard/data/storage/secure_storage.dart';
+
+class ReactivationAuthenticator extends Authenticator {
+  final SecureStorage secureStorage;
+  //final AccountRepository _accountRepository;
+
+  ReactivationAuthenticator(this.secureStorage);
+
+  @override
+  FutureOr<Request?> authenticate(Request request, Response response) async {
+    if (response.statusCode == 401) {
+      final email = await secureStorage.readEmail();
+      final passcode = await secureStorage.readPasscode();
+
+      if (email != null && passcode != null) {
+        /*final either = await _accountRepository.login(email, passcode);
+
+        if (either.isRight) {
+          final Map<String, String> updatedHeaders =
+              Map<String, String>.of(request.headers);
+
+          final newToken = 'Bearer ${either.right.token}';
+          updatedHeaders.update(
+            'Authorization',
+            (String _) => newToken,
+            ifAbsent: () => newToken,
+          );
+          return request.copyWith(headers: updatedHeaders);
+        } */
+      }
+    }
+    return null;
+  }
+}
--- a/lib/utils/reactivation_authenticator.dart
+++ b/lib/utils/reactivation_authenticator.dart
@@ -1,36 +1,90 @@
 import 'dart:async';
 
 import 'package:chopper/chopper.dart';
+import 'package:coffeecard/cubits/authentication/authentication_cubit.dart';
 import 'package:coffeecard/data/repositories/v1/account_repository.dart';
 import 'package:coffeecard/data/storage/secure_storage.dart';
+import 'package:coffeecard/service_locator.dart';
 
 class ReactivationAuthenticator extends Authenticator {
   final SecureStorage secureStorage;
-  //final AccountRepository _accountRepository;
+  DateTime? tokenRefreshedAt;
+  final Duration debounce = const Duration(seconds: 10);
+  int _retryCount = 0;
+  final int _retryLimit = 1;
 
   ReactivationAuthenticator(this.secureStorage);
 
+  bool _canRefreshToken() {
+    if (tokenRefreshedAt == null) {
+      return true;
+    } else {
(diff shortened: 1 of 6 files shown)


lib/utils/reactivation_authenticator.dart
```diff
@@ -0,0 +1,92 @@
...
+    return null;
+  }
+
+  Future<Request?> refreshToken(
+    Request request,
+    Response response,
+  ) async {
+    _retryCount++;
+
+    final accountRepository = sl.get<AccountRepository>();
```
jonasanker, line 60: The service locator should be moved to the repository, else the code is non-testable

lib/utils/reactivation_authenticator.dart
```diff
@@ -0,0 +1,92 @@
...
+  bool _canRefreshToken() {
+    if (tokenRefreshedAt == null) {
+      return true;
+    } else {
+      return tokenRefreshedAt!.difference(DateTime.now()) < debounce;
+    }
+  }
+
+  Future<void> _evict() async {
+    final authenticationCubit = sl.get<AuthenticationCubit>();
```
jonasanker, line 27: The service locator should be moved to the repository, else the code is non-testable

lib/utils/reactivation_authenticator.dart
```diff
@@ -0,0 +1,92 @@
+import 'dart:async';
+
+import 'package:chopper/chopper.dart';
+import 'package:coffeecard/cubits/authentication/authentication_cubit.dart';
+import 'package:coffeecard/data/repositories/v1/account_repository.dart';
+import 'package:coffeecard/data/storage/secure_storage.dart';
+import 'package:coffeecard/service_locator.dart';
+
+class ReactivationAuthenticator extends Authenticator {
```
jonasanker, line 9: Open question: Have you considered the thread-safety of this class? 

I know @TTA777 has introduced parallelism of REST calls which _might_ lead to a scenario where multiple threads end in the authenticator class. This might cause a scenario where read/writes are not visible to the two threads (I think 🤔 )

jonasanker (commented):
@fremartini (Cc @marfavi ) I'm "reopening" the discussion of this implementation as I think there are some design questions which I am unsure has been addressed or have been taken into consideration in the code review :).

Just correct me if my assumptions are wrong 😊 . Let's discuss the question in the comments and see if issues needs to be created :)

lib/utils/reactivation_authenticator.dart
TTA777: So I just tried to find out if secure_storage is intended to be thread safe, and it would appear so. According to this [issue](https://github.com/mogol/flutter_secure_storage/issues/171) only versions before 3.3.5 should have issues, and we are on 5.0.2. With that said, there exists an [open issue](https://github.com/mogol/flutter_secure_storage/issues/354) on secure_storage where it might be unable to read from storage, if the user restored the app from a backup. So it might be prudent to at least wrap our calls to the storage in a try catch as suggested in that issue

lib/utils/reactivation_authenticator.dart
fremartini: i assume you mean that the ReactivationAuthenticator should accept the serviceLocator as parameter and should then retrieve the AccountRepository and AuthenticationCubit from it?

lib/utils/reactivation_authenticator.dart
fremartini: there can arise problems when the two calls on the receiptsPage both return 401 where two attempts to refresh the token are made. I have made some changes here (https://github.com/AnalogIO/coffeecard_app/tree/fix/frem/checkBearerToken) that attempt to address this issue

lib/utils/reactivation_authenticator.dart
jonasanker: Yes that's probably the only solution as we are not able to depend on the account repository with our current service locater 👍🏻

lib/utils/reactivation_authenticator.dart
jonasanker: I assume you are thinking of the static fields you have introduced in the ReactivationAuthenticator?

I understand why you came up with the idea but it's actually only making it worse as static does not guarantee memory visibility across threads (at least in Java). 
Also and maybe more importantly, the Shiftplanning API also uses an Authenticator. Introducing static fields would share the the retry count and debouncer between both API. We don't want that!

That was a little thread safety rant, sorry 😅

I want to bring some good news though. I have tried reading up on the Dart concurrency model. Dart runs **single threaded** by standard in what is called an _isolate_ (memory is isolated and not shared). My understanding is that unless you explicitly create several isolate to make multi threaded code, the code will always be single threaded.

My first thought is therefore that we don't have a thread issue here and no risk of dirty read/writes.

Do you agree on my understanding? @fremartini @TTA777

[Dart Concurrency documentation](https://dart.dev/guides/language/concurrency)

NamanShergill: I would like to point out that dart isolates do not work with platform channels in case that information was missed, i.e. you cannot invoke native methods through isolates. Multi-threading will always have to be handled in the native code of the plugin in that scenario.

fremartini: https://github.com/AnalogIO/coffeecard_app/pull/146

PR #143, opened 2022-02-14
Polishing the receipts page

### LINKED ISSUE #73: (Receipts flow) Display something when there are no receipts to show

### LINKED ISSUE #74: (Receipts flow) ReceiptListEntry with Receipt as its only argument
Instead of having two constructors, `ReceiptListEntry.swipe(productName, time)` and `ReceiptListEntry.purchase(productName, time, quantity, price)`, we should have `ReceiptListEntry(receipt).`

Should explore if we need two Receipt types (one for purchase and one for swipe; maybe subclasses?) or if only one is sufficient.

TTA777:
Added message to receipt page when no receipts exist. Fixed the ordering of receipts, since someone had reversed it. ReceiptListEntry now takes a receipt as argument instead

Closes #74 
Closes #73

### DIFF
--- a/lib/base/strings.dart
+++ b/lib/base/strings.dart
@@ -114,7 +114,7 @@ abstract class Strings {
     return '$price,-';
   }
 
-  // Receipts page
+  // Receipts
   static const receiptsPageTitle = 'Receipts';
   static const singleReceiptPageTitle = 'Receipt';
 
@@ -131,6 +131,10 @@ abstract class Strings {
   static const receiptCardPurchased = 'Purchased';
   static const receiptCardNote = 'This can be found again under Reciepts.';
   static const receiptTapAnywhereToDismiss = 'Tap anywhere to dismiss';
+
+  static const receipts = 'receipts';
+  static String noReceiptsOfType(String noneOfType) =>
+      "You don't have any $noneOfType... yet.";
 
   // Statistics page
   static const statisticsYourStats = 'Your stats';
--- a/lib/base/strings.dart
+++ b/lib/base/strings.dart
@@ -133,8 +133,14 @@ abstract class Strings {
   static const receiptTapAnywhereToDismiss = 'Tap anywhere to dismiss';
 
   static const receipts = 'receipts';
-  static String noReceiptsOfType(String noneOfType) =>
+  static String noReceiptsOfTypeTitle(String noneOfType) =>
       "You don't have any $noneOfType... yet.";
+
+  static const buy = 'buy';
+  static const swipe = 'swipe';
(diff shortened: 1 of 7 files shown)


marfavi:
Added some further polish and bug fixes.

Since #73 had no design spec, I took the liberty of replacing the look of the "no tickets" indicator with something simpler. Moreover, it now shows different messages based on what receipts the user (doesn't) have.

Examples:

User has no receipts of any kind |  User has no purchases, but has swipes
:-------------------------:|:-------------------------:
Any filter | Filtered by purchases
image |  image

Same goes for if the user has no swipes but has some purchases.

Read the commit messages

marfavi:
Otherwise the rest looks good to me. If you approve my changes, I'll approve yours

marfavi (approved):

TTA777 (commented):
Two minor things. Firstly, the for the `hasAnyReceipts` variable in receipts_list_view you end up doing a negation to for no apparent reason. That is, you assign it using the `recipts.isNotEmpty`, but in `_message` you then use the `!` operator. Why not just use the `isEmpty` check to start with, and name the variable accordingly?

Secondly, is the design of the message. Now, I don't really care too much about how it looks, but to me, it seems a little "lonely" for lack of a better term. I.e. It is a very small message, in a very big empty space. Just my two cents.

If you don't agree, feel free to merge it

marfavi:
> Two minor things. Firstly, the for the `hasAnyReceipts` variable in receipts_list_view you end up doing a negation to for no apparent reason. That is, you assign it using the `recipts.isNotEmpty`, but in `_message` you then use the `!` operator. Why not just use the `isEmpty` check to start with, and name the variable accordingly?

Good catch, I fixed it.

> Secondly, is the design of the message. Now, I don't really care too much about how it looks, but to me, it seems a little "lonely" for lack of a better term. I.e. It is a very small message, in a very big empty space. Just my two cents.

I agree, so I tried a bit harder.
How's this look? Still a lot of empty space, which could've been filled out with an illustration, but I couldn't find a good one.

The messages would differ slightly based on the filter you've chosen and what types of receipts you like, just like in my previous comment https://github.com/AnalogIO/coffeecard_app/pull/143#issuecomment-1039467131

image

TTA777:
That looks good to me for now. We can always come back to it at a later date, and maybe get a certain Chair to take a look as well if we feel like getting ideas

marfavi:
Added the design and changed the wording slightly:
`When you $buyOrSwipe tickets, they will show up here.
Go to the Tickets tab to $buyOrSwipe tickets.`;

marfavi:
Branch now up to date with `develop`. Let's get this branch squashed & merged

PR #144, opened 2022-02-14
Added new UI for disabled shop cards for goods, such as 'Buy one drink'

### LINKED ISSUE #129: (Buy single drink/other products) Display ShopCard as disabled with "Coming soon" tooltip

Jens-Fastrup:
Closes #129 

New class "shop_card_disabled" added, which currently holds tooltip "coming soon" which is activated onLongPress. Likewise, colour is now grey to indicate unavailability. The class can also be used for 'Buy others'... etc. 

Couldn't figure out how to add ripple effect, but I guess this is a part of onPressed function, which indeed is left out in the new class (and not necessary). The ripple effect though is mostly style-points to keep it inline with the other boxes. 

Lastly, centering a tooltip is a bit sketchy - but I believe this will work though.

### DIFF
--- a/lib/widgets/components/shop_card_disabled.dart
+++ b/lib/widgets/components/shop_card_disabled.dart
@@ -1,7 +1,4 @@
-import 'dart:ffi';
-
 import 'package:coffeecard/base/style/colors.dart';
-import 'package:coffeecard/widgets/components/helpers/tappable.dart';
 import 'package:flutter/material.dart';
 
 class ShopCardDisabled extends StatelessWidget {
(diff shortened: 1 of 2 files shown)


Jens-Fastrup:
Should I disable 'Buy Syrup, jugs, etc' and Redeem voucher also before pull?

fremartini:
I would disable the buy syrup, jugs etc. and redeem voucher button until they are ready. Furthermore i think it is redundant to add a new widget for the disabled card. I propose to extend the original ShopCard widget with a 'disabled' property with a default value of false. If the widget is disabled, the background color should be something like AppColors.Disabled (if it exists, otherwise create it) and the function should be replaced with the empty function () {}. The tooltip is nice and could also be incorporated to display if the widget is clicked while disabled

PR #145, opened 2022-02-14
Bugfix: Added ".trim()" to ignore whitespaces with email registration

### LINKED ISSUE #126: (Registration flow) Email not trimmed from left
This email should be valid, as any leading and trailing whitespace should be ignored.

image

Jens-Fastrup:
Title says it all 

Closes #126

### DIFF
--- a/lib/cubits/register/register_cubit.dart
+++ b/lib/cubits/register/register_cubit.dart
@@ -9,7 +9,7 @@ class RegisterCubit extends Cubit<RegisterState> {
 
   RegisterCubit({required this.repository}) : super(const RegisterState());
 
-  void setEmail(String email) => emit(state.copyWith(email: email));
+  void setEmail(String email) => emit(state.copyWith(email: email.trim()));
   void setPasscode(String passcode) => emit(state.copyWith(passcode: passcode));
   void setName(String name) => emit(state.copyWith(name: name));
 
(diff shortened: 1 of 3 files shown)


Jens-Fastrup:
I don't quite understand how it fails codecov - can somebody maybe elaborate? :)

marfavi:
> I don't quite understand how it fails codecov - can somebody maybe elaborate? :)

It's because we don't have >= 70 test coverage - don't worry, that's not your fault :)

marfavi (changes requested):
The function `emailIsValid` is expected to validate an already trimmed email.
Trimming should therefore happen in the step before this call.
In this case it is the file `components\entry\register\register_email_text_field.dart` -- should probably be the first step in `_validateEmail`, so we can both save and validate a trimmed version of the email.

Right now, trying to register an account with leading/trailing whitespace results in an "unhandled error" in the final step (that's an issue on the backend, which I presume also expects a trimmed email).

image
image

Jens-Fastrup:
Ah, awesome! Didn't see that we had a class for validating emails! - will look into it

Jens-Fastrup:
Currently ready to be reviewed again, as now trimming is added to UI side aswell as with cubit so - which passes the exception and returns  the trimmed email.

marfavi (changes requested):
Looks good, just fix the one error I pointed out and it's good to go

jonasanker:
Please squash the commits when merging this PR 😊

marfavi (approved):
Looks good. Don't forget to `Squash and merge`!

PR #146, opened 2022-02-15
update how outdated bearer tokens are refreshed

fremartini:
closing #68 

* implemented mutex to control access to functionality that refreshes token based on (https://stackoverflow.com/questions/42071051/dart-how-to-manage-concurrency-in-async-function) and (https://api.dart.dev/stable/1.21.1/dart-async/Completer-class.html)
* _coffeCardChopper and _shiftplanningChopper no longer shares ReactivationAuthenticator


jonasanker:
We must agree on the problem analysis before code reviewing this PR. I think it should be set to draft for now

See discussion in #138

jonasanker (approved):

PR #150, opened 2022-02-18
Add "Connected to test environment" widget

### LINKED ISSUE #106: Indicate that user is in test environment
Widgets can listen to EnvironmentCubit to achieve this functionality.

See discussion in #104

marfavi:
Added a new Scaffold `AppScaffold` that conditionally displays a "Connected to test environment" button.

AppScaffold has two constructors: `AppScaffold.login()` for the login page and `AppScaffold()` for everything else.

closes #106


fremartini:
@marfavi i made some changes, please review

lib/widgets/custom_scaffold.dart
```diff
@@ -0,0 +1,145 @@
+import 'package:coffeecard/base/strings_environment.dart';
```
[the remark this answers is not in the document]
marfavi: Why not just rename it to `Scaffold` then?

lib/widgets/app_scaffold.dart
```diff
@@ -0,0 +1,135 @@
...
+import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:gap/gap.dart';
+
+class AppScaffold extends StatelessWidget {
+  final Widget? title;
+  final Widget body;
+  final bool resizeToAvoidBottomInset;
+
+  AppScaffold({
```
marfavi, line 15: I still want two constructors here, this is more unclear than what we had before

lib/widgets/custom_scaffold.dart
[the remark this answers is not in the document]
fremartini: then it clashes with the standard Scaffold, that seems like a bad idea

lib/widgets/app_scaffold.dart
fremartini: I don't agree with having the scaffold be in either 'normal' or 'login' mode, this widget should not be that specific and not writing specific code that only a single page will use, i much prefer taking the general approach as i have done here and making the title nullable and determine the behaviour of the widget on wheter this is null or not. I agree that checking if the String title is null maybe be additional code, but  then i propose to remove the ```String? title``` field and adding ```this.title``` instead. Then we can simply check if the title widget is given or not and we can reuse this other places where a ttitle may not be required

fremartini: with that said i may have an idea, stay tuned

lib/widgets/custom_scaffold.dart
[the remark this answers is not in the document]
marfavi: We already have `lib/widgets/components/card.dart`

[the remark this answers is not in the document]
fremartini: fair enough, can you change it?

lib/widgets/app_scaffold.dart
fremartini: i pushed a new constructor, can you review and resolve this if you agree?

lib/widgets/components/scaffold.dart
```diff
@@ -0,0 +1,139 @@
...
+      ),
+      body: BlocBuilder<EnvironmentCubit, Environment>(
+        builder: (context, state) {
+          return Column(
+            crossAxisAlignment: CrossAxisAlignment.stretch,
+            children: [
+              if (state.isTest) const _EnvironmentBanner(tappable: true),
+              Expanded(
+                child: Container(
+                  color: (title != null) ? AppColor.background : null,
```
marfavi, line 50: The new constructor `Scaffold.noTitle` does not imply that the background of the body will change.
The name `Scaffold.login` fits better in this scenario

fremartini: i see, i still don't agree with having a constructor fitted to a specific page. I propose to add an optional background field in the noTitle constructor to preserve the reuseable aspect, this way the loginPage can control the color of the background. Thoughts?

marfavi: Check the new changes

fremartini (approved):
LGTM

fremartini (approved):

PR #153, opened 2022-02-20
Update Flutter version

marfavi:


jonasanker (changes requested):
@marfavi Can you please check if Dart is also updated?

Also, please update both pipelines files to use the new Flutter version 😊

marfavi:
After merging, we should all run `flutter upgrade` and `flutter pub get` :)

jonasanker (approved):
Approved ✔️ 

@marfavi Will you write in Slack and ask everyone to update their flutter version?

PR #157, opened 2022-02-20
Move user information to cubit

### LINKED ISSUE #119: Encapsulate user information in cubit
The user information is needed on the tickets, statistics and settings-pages. This information should be made accessible once the user has logged in

### LINKED ISSUE #140: display user stats retrieved from user data cubit
The current users statistics are being retrieved from the API. This is unnecessary as the information is already available in the app. Blocked by #119

fremartini:
closes #119 
closes #140 

* Add UserCubit to manage currently logged in user
* Statistics page fetches ranks from UserCubit
* SettingsPage fetches ID from UserCubit


jonasanker:
Is someone already reviewing this?

lib/widgets/pages/settings/your_profile_page.dart
```diff
@@ -34,6 +34,8 @@ class _EditProfile extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final programmeId = context.read<UserCubit>().state.user?.programmeId;
```
marfavi, line 37: Use a BlocBuilder

lib/cubits/statistics/statistics_state.dart
```diff
@@ -33,21 +33,16 @@ extension DropdownName on StatisticsFilterCategory {
 class StatisticsState extends Equatable {
```
marfavi, line 33: Should look at refactoring this state class to multiple classes now that it has been simplified

lib/cubits/statistics/statistics_cubit.dart
```diff
@@ -10,10 +9,8 @@ part 'statistics_state.dart';
 
 class StatisticsCubit extends Cubit<StatisticsState> {
   final LeaderboardRepository leaderboardRepository;
-  final AccountRepository accountRepository;
 
-  StatisticsCubit(this.leaderboardRepository, this.accountRepository)
-      : super(StatisticsState());
+  StatisticsCubit(this.leaderboardRepository) : super(StatisticsState());
 
   void filterStatistics(StatisticsFilterCategory filterBy) {
```
marfavi, line 15: I don't understand this method. Can you explain what it does?

marfavi (changes requested):
## Notes

Wrong usage of `context.read`.
`read` will not trigger a rebuild when the cubit state changes.
The correct method here is to use `context.watch`.
I'd strongly prefer to use a `BlocBuilder` instead of `context.watch`, though

On for `BlocBuilder` vs `context.watch`:
https://github.com/felangel/bloc/issues/1902#issuecomment-723664125

---

Remove DropdownName.preset

The `preset` getter on DropdownName is seeing usage outside of its intended use case (LeaderboardRepository.getLeaderboard).
Frankly it shouldn't even exist for this use case either (the repository could take in a `StatisticsFilterCategory` and do the conversion from enum to int in `LeaderboardRepository.getLeaderboard`.

Discuss

lib/cubits/statistics/statistics_cubit.dart
fremartini: sure.

1. if a refresh is triggered with the same filter, ignore
2. the dropdown has changed, change the cubit state to the new one
3. fetchLeaderboards() is called where it fetches data with the current selected preset (stored in the cubit)

fremartini:
@marfavi overall i like the changes. Is there anything in particular you would like to discuss? otherwise i suggest we merge this into develop

marfavi (approved):
Remember to squash and merge

PR #171, opened 2022-02-21
MobilePay spike

### LINKED ISSUE #38: Shop/Buy tickets flow
- [x] #43
- [x] #81

### LINKED ISSUE #43: MobilePay support
Create support for MobilePay in Flutter

Tasks:
- Create an abstract class for making a MobilePay payment

Implement class
- [x] #44
- [x] #45 

Relevant documentation [MobilePay Dev: Build on Mobile](https://mobilepaydev.github.io/MobilePay-Payments-API/docs/guides/app-payments/build-on-mobile)

Link to backend issue https://github.com/AnalogIO/analog-core/issues/31

### LINKED ISSUE #44: MobilePay iOS

### LINKED ISSUE #81: Migrate legacy mobilepay to new approach
The current MP implementation follows legacy practices. Move the implementation to the approach 
https://mobilepaydev.github.io/MobilePay-Payments-API/docs/guides/app-payments/build-on-mobile

TTA777:
Closes #38 
Closes #43 
Closes #44, Well it should, don't have the capabilities to test it :/
Closes #81 

### Can complete purchases using the MobilePay service.
- Uses deeplink to navigate to mobilepay, and can recevieve a deeplink from MobilePay
- Detects the app is back in the foreground, and checks with analog-core that the purchase has completed
- Redirects the user back to the home screen and show a receipt
- Retrive a users tickets, and refresh these when they have made a purchase
- Refreshes a users receipts when they have made a purhcase

### What it can't do
- The deeplink from mobilepay is currently only used to get us back in the app, we don't use the parameter/ path it provides us with
- The backend check for a purchases' status always returns completed,  the app instead looks at the mobilePay status included in the response from the backend. This will likely have to be changed once the backend supports it
- Needs more UI polish


TTA777:
If anyone wants to test the entire mobilepay flow, they need to install the mobile-pay sandbox app as described here https://mobilepaydev.github.io/MobilePay-Payments-API/docs/testing

fremartini:
@TTA777 this branch has merge conflicts

TTA777:
@fremartini Only cause develop kept getting updated, it's fixed for now.

fremartini:
> @fremartini Only cause develop kept getting updated, it's fixed for now.

awesome. Will review tomorrow :)

lib/payment/mobilepay_service.dart
```diff
@@ -1,121 +1,103 @@
...
+
+      return Right(
+        Payment(
+          id: purchaseResponse.id!,
+          paymentId: paymentDetails.paymentId!,
+          status: PaymentStatus.awaitingPayment,
+          deeplink: paymentDetails.mobilePayAppRedirectUri!,
+          purchaseTime: purchaseResponse.dateCreated!,
+          price: purchaseResponse.totalAmount!,
+          productName: 'Placeholder',
```
fremartini, line 34: do we have a solution to this? Where is this used?

lib/widgets/components/purchase/purchase_process.dart
```diff
@@ -0,0 +1,116 @@
...
+              }
+            },
+          ),
+        ],
+      ),
+    );
+  }
+
+  StatelessWidget makeDialog(String title, {Widget? content}) {
+    return content != null
```
fremartini, line 85: What is the difference between AlertDialog and SimpleDialog? seems like there is a bit of duplication here

lib/widgets/components/purchase/purchase_process.dart
```diff
@@ -0,0 +1,116 @@
...
+                return makeDialog(Strings.purchaseTalking);
+              } else if (state is PurchaseProcessing ||
+                  state is PurchaseStarted) {
+                return makeDialog(Strings.purchaseTalking);
+              } else if (state is PurchaseVerifying) {
+                return makeDialog(Strings.purchaseCompleting);
+              } else if (state is PurchaseCompleted) {
+                return makeDialog(Strings.purchaseSuccess);
+              } else if (state is PurchasePaymentRejected) {
+                return makeDialog(
```
fremartini, line 60: How do we add the 'tap anywhere to dismiss' text to the rejected popup? seems like we may be able to reuse the same widget that the receipt card uses somehow?

lib/payment/mobilepay_service.dart
TTA777: The current solution is to not use it. The "fix" resides in the purchase overlay where it is using the name of the product that was passed to it for showing the receipt to the customer, instead of using the one provided by the payment. Personally I would prefer getting the product name from the backend, but it is not really a necessity, as I can get it from the other place

lib/widgets/components/purchase/purchase_process.dart
TTA777: AlertDialog takes an array of widgets representing the actions that a user have, e.g. the "Ok" button. Since these are actually nullable, we could likely reuse the AlertDialog for both cases, without any issue, it just seemed to better communicate their intended purpose by using the widget designed for those cases

lib/widgets/components/purchase/purchase_process.dart
TTA777: It was actually on purpose that it had not been added, since the user can't press anywhere to dismiss it. They can only dismiss it by pressing the "Ok". The reasoning being that we want them to acknowledge that an error happened, and not accidentally dismiss the dialogue

lib/widgets/components/purchase/purchase_process.dart
fremartini: allright. I suggest moving the Text(title) and shape into shared variables then

lib/payment/mobilepay_service.dart
fremartini: is productName required, if not could we omit it then?

TTA777: Currently, it is required, but we could make it nullable if there is a preference. We could also remove it entirely from the object, until we are able to get the information from the backend. I would personally prefer removing it, to making it nullable

lib/widgets/components/purchase/purchase_process.dart
TTA777: Sure, but you are then fine with it being two separate types of widget? I thought about it for a bit more, and I don't really like how I distinguish between which widget gets created, i.e. I assume we want an alert if content is provided. Any thoughts?

lib/widgets/components/purchase/purchase_overlay.dart
```diff
@@ -0,0 +1,81 @@
...
+              listener: (context, state) async {
+                if (state is PurchaseCompleted) {
+                  hide(); //Removes this overlay
+                  final navigator = Navigator.of(context);
+                  //TODO consider using popUntil instead of this
+                  while (navigator.canPop()) {
+                    //Gets the user back to the home-screen of the app
+                    navigator.pop();
+                  }
+                  //TODO Consider if these calls should be moved elsewhere, e.g. inside the purchase cubit
```
TTA777, line 46: Personally, I don't really think the calls to refresh the tickets, and receipts belong here, but not sure where else they make sense to have. Thoughts? If we keep them here, we should move the cubits to the constructor for testability, at the very least.

lib/payment/mobilepay_service.dart
fremartini: i think we should remove it for now and introduce it once it is needed

lib/widgets/components/purchase/purchase_overlay.dart
fremartini: I think this is fine for now. We should probably do something smarter once we get to testing our logic but for now i would prefer to get this into dev :)

lib/payment/mobilepay_service.dart
TTA777: Removed

lib/widgets/components/purchase/purchase_overlay.dart
TTA777: Alright, in that case I will just leave the TODO for now, so we can get back to it later

fremartini (dismissed):

lib/widgets/components/purchase/purchase_process.dart
TTA777: Good idea, just pushed a change to do so. Take a look and confirm whether that is what you imagined. If everything looks alright with you, feel free to merge it in.

fremartini (approved):

PR #173, opened 2022-02-23
Add continue buttons to register flow pages

### LINKED ISSUE #93: Add continue buttons in register flow

fremartini:
* each section of the register flow has a continue button that acts in accordance with already defined behavior regarding errors
* rounded_button now hides splash effect when disabled

closes #93

### DIFF
--- a/lib/widgets/components/entry/register/register_continue_button.dart
+++ b/lib/widgets/components/entry/register/register_continue_button.dart
@@ -1,7 +1,7 @@
 import 'package:coffeecard/widgets/components/rounded_button.dart';
 import 'package:flutter/material.dart';
 
-class RegisterContinueButton extends StatefulWidget {
+class RegisterContinueButton extends StatelessWidget {
   final Function() onPressed;
   final bool enabled;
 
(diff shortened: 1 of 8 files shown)


lib/widgets/components/entry/register/register_continue_button.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:coffeecard/widgets/components/rounded_button.dart';
+import 'package:flutter/material.dart';
+
+class RegisterContinueButton extends StatefulWidget {
```
TTA777, line 4: I don't fully understand the need of the button being a stateful widget. Could you elaborate why this is necessary, or if it isn't make it stateless instead?

TTA777 (changes requested):
One comment I would like addressed. Rest likes fine in my eyes

lib/widgets/components/entry/register/register_continue_button.dart
fremartini: its not, i forgot to remove it :) fixing

TTA777 (approved):
LGTM, let's get this merged in. If someone else wants to fine-tune the appearance, we can always do so later

PR #175, opened 2022-03-05
add 'show me on leaderboard button' to settingspage

### LINKED ISSUE #162: Allow user to set itself as Anonymous
A user should be able to mark himself to be anonymous on the leaderboards. This is already implemented in the existing apps.
Setting should be added to Account Settings.

fremartini:
closes #162 

displays a switch on the settingspage allowing the logged in user to toggle their 'privacyActivated' field

@jonasanker does the backend omit people with this field set to 'true' or should it be handled in the frontend? I do believe it would be better to handle in the backend as we only receive names and scores from the API call. Thoughts?

### DIFF
--- a/lib/base/strings.dart
+++ b/lib/base/strings.dart
@@ -190,7 +190,10 @@ abstract class Strings {
   static const change = 'Change';
   static const logOut = 'Log out';
   static const deleteAccount = 'Delete account';
-  static const showOnLeaderboard = 'Show me on the leaderboard';
+  static const appearAnonymous = 'Appear anonymous in leaderboard';
+  static const appearAnonymousSmall = 'Appear anonymous';
+  // FIXME decide between these or:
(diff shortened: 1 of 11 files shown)


jonasanker:
@fremartini When a user has set privacy to true, they are either omitted or listed as anonymous on the leaderboard. I can not remember which it is but the frontend shouldn't need to do anything

fremartini:
> @fremartini When a user has set privacy to true, the backend either omits or lists the as anonymous on the leaderboard. I can not remember which it is but the frontend shouldn't need to do anything 

Cool, then i think this feature is done

marfavi (changes requested):
Check #180 for suggested changes

fremartini:
@jonasanker @marfavi please review

marfavi (approved):

Issue #152, opened 2022-02-20
Display invalid-ness of receipt


marfavi:
- [x] #159
- [x] #158
- [x] #160

Receipts should have an indicator that displays if it is invalid.

Receipts are displayed as invalid if the user is connected to a test environment.

PR #179, opened 2022-03-06
Test (environment) warning on receipt by change of color

### LINKED ISSUE #158: Add environment warning on receipt-popup

Jens-Fastrup:
closes #158 

- Color on receipt change to yellow if environment is test. 
- The color yellow is just a placeholder, do not recall seeing any for a test environment - so could be subject to change.

 - Is this the use of state/cubit correct way to do it in Dart? 
- Should we have a color case for Environment.unknown?


Jens-Fastrup:
- I think my main concern is how one would go around using Environment.unknown in _receipt_overlay.dart_. Best case it ReceiptCard should not be require it, but currently it is set to unknown as it has no use in the class. Does anybody have an idea to remove environment with this class? likewise adding a blocbuilder would be overkill and have no use other than getting the right state.

TTA777:
> * I think my main concern is how one would avoid using **Environment.unknown** in _receipt_overlay.dart_. Best case ReceiptCard should not be require it, but currently it is set to **unknown** as it should have no use in the class. Does anybody have an idea to remove the environment from this class? likewise adding a blocbuilder would be overkill and have no use other than getting the right state.

ReceiptCard should certainly have the environment as a required field as you have done, since we would want it to display the correct colour, no matter where we use the receiptCard. With that said, we want to somehow provide the current environment to the card inside the overlay. Otherwise, when a user swipes a ticket, or performs a purchase, the environment would not be reflected on the recieptOverlay. 

For how you get the correct environment to the card in the overlay, then you could either do it with a blockBuilder as you suggested yourself. Alternatively, since we don't really need it to react to changes in the state as the blockbuilder does, you could do a context.read<EnvironmentCubit>() which would give you access to the cubit, and its state. Usually that is only used for making function calls to the cubit, but since we don't really want to react to changes in this instance, it should be fine.

If you run into an issue where you can't find the cubit in the current context, you might need to register it in the service locator instead, and get it from there

TTA777 (changes requested):
Ohh, and on the code itself, that looks good to me. Considering the state itself here is just an enum, I think you are using it correctly. Not sure if I agree with the person who made it an enum, but that's another matter. Just address the overlay, passing the correct state to the card, and it should be fine.

marfavi:
Environment type will never be unknown when app has left the splash screen

marfavi:
See how the code would look if Environment was refactored into a state class instead of an enum #181

Jens-Fastrup:
I agree that making Environment a class instead of an enum should be the way. This way, as Omid has proposed, overlay would not have to have a state like unknown predefined. Thanks for feedback Thomas and Omid - will look into the suggestions tonight.

marfavi:
I think the issue this PR would close is #158 -- not #152 

Let's talk design:

* **Regarding the color:** Try #F5FF68 -- it's a (very slightly) less saturated yellow :)

* **Regarding the Receipt pop-up:** Receipt pop-up refers to the receipt shown in an overlay (with dimmed background). Here, no "test environment" warning is visible (its dimmed out in the background). A change in receipt color is fine, but I would tell the user why the card is yellow as well (sketch):

  image

* **Regarding the Receipt card**: Receipt card refers to the receipt shown in the Receipts flow. Here the "test environment" warning is visible under the app bar. The card can "afford" to not be yellow, as the environment warning is already visible. I don't mind the extra indication though.

Otherwise, looks good!

Jens-Fastrup:
- Yea, adding it to the Receipt pop-up would be useful too I see. But when I want to test it by buying tickets for Test Coffee, I get redirected to MP and nothing else - can't get the pop up running? Didn't it work before, did I miss something? I'm develop branch and pulled.
- I kinda like the receipt color consistency depending on the environment, that way, nobody can trick our awesome baristas with a quick show of their phone 😅

TTA777:
@Jens-Fastrup  I don't think the develop branch currently shows the receipt pop-up. I added that to the mobilepay spike branch at least, so I think it is unused in develop. Just to test it, you could change it, so it triggers the receipt pop-up when you press to make a purchase. Just don't commit that change, and it should be fine. 

Regarding colours, I'm not gonna comment much, other than to say it should be clear to everyone who sees it that something is wrong. So unlike normally where we want everything to be pretty, we probably want something extremely garish that is shouting to even the casual observer that something is not right.

TTA777:
@Jens-Fastrup Not sure what the status is here, but if you rebase on develop, then you'll get functionality for both purchases through mobilepay, and swiping of tickets if that helps you out with testing. In any event you'll need to do so since there are now merge conflicts

Jens-Fastrup:
Perfect! I'm off tomorrow, so will have a look at it 🙏

tor. 10. mar. 2022 19.15 skrev Thomas Andersen ***@***.***>:

marfavi:
Closing - see #205

PR #180, opened 2022-03-07
Suggested changes to anonymous user selection

marfavi:
moved location of setting to the Your Profile view
added loading indicator when toggling between on/off
text of setting changed to "Appear anonymous in leaderboard" (or "Appear anonymous" on iPhone 5-sized screens)
added description for setting group

### DIFF
--- a/lib/base/strings.dart
+++ b/lib/base/strings.dart
@@ -199,8 +199,6 @@ abstract class Strings {
   static const deleteAccount = 'Delete account';
   static const appearAnonymous = 'Appear anonymous in leaderboard';
   static const appearAnonymousSmall = 'Appear anonymous';
-  // FIXME decide between these or:
-  // static const appearAnonymous = 'Hide my info in leaderboard';
 
   static const signInWithFingerprint = 'Sign in with fingerprint';
 
--- a/lib/base/strings.dart
+++ b/lib/base/strings.dart
@@ -197,7 +197,10 @@ abstract class Strings {
   static const change = 'Change';
   static const logOut = 'Log out';
   static const deleteAccount = 'Delete account';
-  static const showOnLeaderboard = 'Show me on the leaderboard';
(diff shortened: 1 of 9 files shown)


lib/base/strings.dart
```diff
@@ -197,7 +197,8 @@ abstract class Strings {
   static const change = 'Change';
   static const logOut = 'Log out';
   static const deleteAccount = 'Delete account';
-  static const showOnLeaderboard = 'Show me on the leaderboard';
+  static const appearAnonymous = 'Appear anonymous in leaderboard';
```
fremartini, line 200: should be 'on' leaderboard

fremartini (changes requested):

lib/widgets/pages/settings/your_profile_page.dart
```diff
@@ -18,80 +19,85 @@ class YourProfilePage extends StatelessWidget {
...
+              valueWidget: const SettingDescription(),
+              onTap: () {},
+            ),
+            SettingListEntry(
+              name: deviceIsSmall(context)
+                  ? Strings.appearAnonymousSmall
+                  : Strings.appearAnonymous,
+              valueWidget: Switch(
+                value: user.privacyActivated,
+                onChanged: (privacyActived) async {
```
fremartini, line 91: can you verify this value is the expected one? as in true when toggled on and false when off

marfavi:
> image should this be centered?

No, see pictures from design spec
image
image
image

lib/widgets/pages/settings/your_profile_page.dart
marfavi: looks correct to me

lib/base/strings.dart
marfavi: fixed in new commit, also changed it in the settings group description

PR #181, opened 2022-03-07
suggested change to Environment warning

marfavi:
small write up to demonstrate how the environment state would look like if it was a state class instead of an enum

unknown state has been replaced by EnvironmentInitial/EnvironmentError classes that have to be checked

personally i think either implementation looks OK and merging this branch should be optional


Jens-Fastrup:
I like the abstract class implementation, nice and clean - also makes overlay more intuitive 🙌

PR #186, opened 2022-03-10
Display programme instead of ID

### LINKED ISSUE #184: Display programme name instead of its ID
The settings page currently displays the users programme ID instead of the corresponding text. We need to use the API to lookup the name, or hardcode a table into the app as programmes are very unlikely to ever change

fremartini:
closes #184

Retrieves programme info from backend and displays
* 'shortName' in the occupation settings field
* 'fullName' on the usercard


lib/widgets/components/stats/stat_card.dart
```diff
@@ -45,3 +44,19 @@ class StatisticsCard extends CardBase {
           ),
         );
 }
+
+String formatLeaderboardPostfix(int rank) {
+  final rankStr = rank.toString();
```
TTA777, line 49: No need for string conversion, you can use % 10 instead

lib/widgets/components/stats/stat_card.dart
```diff
@@ -45,3 +44,19 @@ class StatisticsCard extends CardBase {
...
+  final lastDigit = rankStr[rankStr.length - 1];
+
+  switch (lastDigit) {
```
TTA777, line 52: This will give wrong resutls for 11, 12, and 13. I suggest just having a check before for the range 11 - 19

lib/cubits/user/user_cubit.dart
```diff
@@ -1,22 +1,38 @@
...
   Future<void> fetchUserDetails() async {
     emit(UserLoading());
 
     final either = await accountRepository.getUser();
 
     if (either.isRight) {
-      emit(UserLoaded(either.right));
+      var user = either.right;
+
+      final programmes = await programmeRepository.getProgramme();
```
TTA777, line 24: Any way that we could save this data somewhere, in order to reuse it for when a user wants to update their current programme?

TTA777 (changes requested):
formatLeaderboardPostfix needs fixing

lib/cubits/user/user_cubit.dart
fremartini: yes that we should be able to handle. Opened #187

TTA777 (approved):
lgtm, though I would prefer `def` being called `default` :P

fremartini:
> lgtm, though I would prefer `def` being called `default` :P 

Yes, 'default' is however a keyword in Dart so i dont think its possible. Hence why 'def'

PR #188, opened 2022-03-12
Receipts displays local time

### LINKED ISSUE #183: Receipt displays wrong time
Swipe happened at 17:27
image

fremartini:
closes #183 

The data objects still contain their UTC time, but the local time is displayed. Would it be preferred to save the local time in the objects instead? Thoughts?

### DIFF
--- a/lib/widgets/components/receipt/receipt_card.dart
+++ b/lib/widgets/components/receipt/receipt_card.dart
@@ -35,8 +35,14 @@ class ReceiptCard extends CardBase with IgnorePointerCard {
               const Gap(16),
               Text(productName, style: AppTextStyle.ownedTicket),
               const Gap(12),
-              Text(timeSince(time), style: AppTextStyle.textFieldBold),
-              Text(_formatter.format(time), style: AppTextStyle.textField),
+              Text(
+                timeSince(time.toLocal()),
+                style: AppTextStyle.textFieldBold,
+              ),
+              Text(
+                _formatter.format(time.toLocal()),
+                style: AppTextStyle.textField,
+              ),
             ],
           ),
           gap: 120,


lib/widgets/components/receipt/receipt_card.dart
```diff
@@ -35,8 +35,8 @@ class ReceiptCard extends CardBase with IgnorePointerCard {
               const Gap(16),
               Text(productName, style: AppTextStyle.ownedTicket),
               const Gap(12),
-              Text(timeSince(time), style: AppTextStyle.textFieldBold),
-              Text(_formatter.format(time), style: AppTextStyle.textField),
+              Text(timeSince(time.toLocal()), style: AppTextStyle.textFieldBold),
```
TTA777: Wouldn't we have a potential issue if a user's phone was set to another timezone? At least, I presume it is just checking the phones' timezone, and not going off position. Any way to enforce that it is Denmark's current time, i.e. choose between CET and CEST as applicable?

TTA777 (commented):
I would prefer saving times in UTC, always, and then just make the conversion when we need to display it. Just one concern, see other comment

lib/widgets/components/receipt/receipt_card.dart
fremartini: then the receipts would be displayed in the phones' timezone. This may be a potential problem but i don't see why people would have another time selected while studying in Denmark :D should we handle this? currently i don't see an easy way to do this

TTA777: Thought it was worth considering, due to the relatively high amount of international students/ faculty. It does seem dart has quite poor support for time conversions by default. One option would be to create an extension on DateTime class which could do the conversion by manually adding as appropriate, or use a package from pub.dev. That is, if we consider it enough of an issue.

fremartini: Opened #191

fremartini: I propose we leave this for now and handle it when we have launched the MVP

TTA777: Fine by me

TTA777 (approved):

Issue #163, opened 2022-02-20
Show Analog opening hours in Settings page


jonasanker:
Show Analog opening hours _somewhere_ in the Settings page

PR #206, opened 2022-03-14
remove opening hours from settings; add disabled option to SettingsListEntry

fremartini:
#163

### DIFF
--- a/lib/widgets/components/settings_list_entry.dart
+++ b/lib/widgets/components/settings_list_entry.dart
@@ -7,10 +7,12 @@ class SettingListEntry extends StatelessWidget {
   final String name;
   final Widget? valueWidget;
   final bool destructive;
+  final bool disabled;
   final void Function()? onTap;
 
   const SettingListEntry({
     required this.name,
+    this.disabled = false,
     this.onTap,
     this.valueWidget,
     this.destructive = false,
@@ -21,7 +23,8 @@ class SettingListEntry extends StatelessWidget {
     return SizedBox(
       height: 50,
       child: ListEntry(
-        onTap: onTap,
+        backgroundColor: disabled ? AppColor.gray : null,
+        onTap: disabled ? null : onTap,
         leftWidget: !destructive
             ? Text(name)
             : Text(name, style: const TextStyle(color: AppColor.error)),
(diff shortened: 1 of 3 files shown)


marfavi (changes requested):
Widgets in the material library determine enabled/disabled state based on the presence of the `onTap` field, I suggest we do the same.

You can write a bool getter `bool get disabled => onTap == null;` to determine if the button should be disabled instead of having a `disabled` field.

fremartini:
> Widgets in the material library determine enabled/disabled state based on the presence of the `onTap` field, I suggest we do the same.
> 
> You can write a bool getter `bool get disabled => onTap == null;` to determine if the button should be disabled instead of having a `disabled` field.

thats a nice way of handling it. Using this idea the 'Appear anonymous' button gets grayed out as it does not use the 'onTap' field

image

```dart
  SettingListEntry(
    name: ...
    valueWidget: Switch(
      value: user.privacyActivated,
      onChanged: (privacyActivated) async {
        await context
            .read<UserCubit>()
            .setUserPrivacy(privacyActived: privacyActivated);
      },
    ),
  ),
```

marfavi:
> > Widgets in the material library determine enabled/disabled state based on the presence of the `onTap` field, I suggest we do the same.
> > 
> > You can write a bool getter `bool get disabled => onTap == null;` to determine if the button should be disabled instead of having a `disabled` field.
> 
> That could be a nice way of handling it. However using this idea the 'Appear anonymous' button gets grayed out as it does not use the 'onTap' field
> 
> image
> 
> ```dart
>   SettingListEntry(
>     name: ...
>     valueWidget: Switch(
>       value: user.privacyActivated,
>       onChanged: (privacyActivated) async {
>         await context
>             .read<UserCubit>()
>             .setUserPrivacy(privacyActived: privacyActivated);
>       },
>     ),
>   ),
> ```

It should use the onTap. (The setting should toggle when tapping anywhere on the list entry, not just the switch)

marfavi (approved):
OK, but let's return opening hours in the future.

fremartini:
> OK, but let's return opening hours in the future.

yes, #163 addresses this

PR #215, opened 2022-03-17
Entry flow is responsive

### LINKED ISSUE #178: Login page overflows
The login page has a small overflow when a presumably too large keyboard is present

marfavi:
Resolves #178

No page in login or register flow should overflow.

- Analog logo becomes smaller based on screen width
- Numpad becomes smaller based on screen width
  - Numpad paddings adjusted
- Login pages no longer vertically centered (now matches design spec)
  - Login page padding from top based on screen width
- All login and register pages made scrollable as a last resort

Future:
- Consider removing overscroll or replace it with bouncing (design desicion)

### DIFF
--- a/lib/widgets/analog_logo.dart
+++ b/lib/widgets/analog_logo.dart
@@ -1,13 +1,15 @@
+import 'package:coffeecard/utils/responsive.dart';
 import 'package:flutter/material.dart';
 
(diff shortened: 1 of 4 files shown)


fremartini (approved):
LGMT. No longer overflows on device that previously caused problems

PR #220, opened 2022-03-18
Add support for changing occupation; Change how navigation works

### LINKED ISSUE #187: Allow user to update their programme

TTA777:
Closes #187 

The first commit is an opinionated refactor of the settings page, to make it easier to read the code in my opinion.
Second one is the addition of changing a user's programme. This serves as a PoC for how it could be done, since it was not something considered in the initial design, so more styling is needed


marfavi (changes requested):
Check #221 for suggested changes. In particular, the way pages are pushed to the stack I think needs to change.

marfavi (approved):

PR #221, opened 2022-03-18
Change how navigation and purchase works

marfavi:
## navigation
* `TicketsFlow`, `ReceiptsFlow`, etc removed
* `AppFlow` added
* navigation will use the classic `Navigator.push()` and similar methods - but we have to use them correctly
* added static `route` getter for every page that will be pushed to a nav stack
  * pages with parameters in their constructor has a `routeWith({parameters})` method instead

## purchase
* receipts update after successful purchase

* now utilizing "route results": popping routes off the nav stack with return values attached.
  * this method removes the need for a lot of state.

* `PurchaseOverlay` removed
* `PurchaseProcess` removed
* `PurchaseCubit` removed

* `BuyTicketBottomModalSheet` removed
  * tapping a `BuyTicketsCard` immediately starts purchase

* references to Apple Pay removed
* `MobilePayService` no longer implements `PaymentHandler`
  * `PaymentHandler` removed

* `BuyTicketsCard` now in charge of both starting & handling purchase process

* `purchase.dart` added with the most of the deleted functionality added back in a single file

* "loading dialog with status" replaced by loading indicator with no status
  * "Working on it..." text shows up after two seconds of loading

---

### after swiping:
* receipts update
* stats update

### service locator:
* remove `TicketsCubit` & `ReceiptCubit` from get_it

### old dead files removed
* `popup_card.dart` (replaced by receipt pop-up long ago)
* `single_coffee_card.dart` (not in use)
* `.keep` in lib\payment (no longer needed)

tidy up


lib/widgets/pages/settings/change_email_page.dart
```diff
@@ -0,0 +1,54 @@
...
+            context.read<UserCubit>().setUserEmail(email);
+            await appDialog(
+              context: context,
+              title: 'Email changed!',
+              children: [
+                const Text('Please log in again with your updated email.'),
+              ],
+              actions: [
+                TextButton(
+                  onPressed: () => closeAppDialog(context),
```
TTA777, line 40: Why are we using a stateful widget to get access to mounted, instead of simply logging the user out when pressing the "OK" button?

lib/widgets/pages/settings/change_occupation_page.dart
```diff
@@ -0,0 +1,45 @@
...
+                return SettingListEntry(
+                  name: '${programme.fullName!} (${programme.shortName!})',
+                  valueWidget: Radio(
+                    value: programme.shortName!,
+                    groupValue: state.user.programme.shortName,
+                    onChanged: (_) {},
+                  ),
+                  onTap: () {
+                    context.read<UserCubit>().setUserProgramme(programme.id!);
+                    // Navigator.pop(context);
```
TTA777, line 33: Why is this commented out?

lib/widgets/routers/settings_flow.dart
```diff
@@ -1,16 +1,16 @@
...
 class SettingsFlow extends StatelessWidget {
   static Route get route => MaterialPageRoute(builder: (_) => SettingsFlow());
 
-  static const settingsRoute = '/';
-  static const yourProfileRoute = '/your-profile';
-
   static final navigatorKey = GlobalKey<NavigatorState>();
-  static void push(String route) => navigatorKey.currentState!.pushNamed(route);
+
+  static void push(Widget page) {
```
TTA777, line 9: I generally like this change to SettingsFlow, but maybe we should stop to consider the purpose of it? I never quite understood the reasoning for the old version of this, and I presume @fremartini didn't do so either, since I think he was the one who originally used the navigator to push to the stack, instead of the flow classes. 

While the new version is more straight forward, and easier to use, I don't see how it differs from the page_pusher that I introduced as a convenience method. If it is only supposed to be a convenience method, why are we even using multiple different ones? And if that is not, its only purpose, what else do you see as its purpose?

lib/widgets/pages/settings/change_email_page.dart
marfavi: I want the user logged out only after dismissing the app dialog, hence the await keyword when calling appDialog.

When accessing the context across asynchronous gaps, the mounted check is needed

lib/widgets/pages/settings/change_occupation_page.dart
marfavi: Should be removed.

Exiting a page after choosing a value from a list of values is unexpected behavior.

marfavi: A modal would be more suitable if we want to close it after choosing a value.

lib/widgets/routers/settings_flow.dart
marfavi: OK i'm falling down the rabbit hole trying to come up with an elegant alternative, so let's discuss this in person

TTA777 (changes requested):
This should be split into two separate pull requests. The navigator changes are looking promising, but I did not look too closely, since there are too many things in this pull request. I will likely disagree strongly with a lot of the changes to the purchase logic, as it currently stands, but let's get to that once that is its own PR

PR #224, opened 2022-03-19
remove abstract modifier from CardBase

### LINKED ISSUE #212: `CardBase` should not be a class to be inherited
Currently it stops us from doing some specific things with some cards (no access to a `BuildContext` and therefore no responsive features on the card's properties)
Use of composition should be utilized instead.

fremartini:
closes #212 

* CardBase is no longer an abstract class
* Cards that previously inherited from CardBase, now returns an instance of it with appropriate styling


marfavi (approved):
Looks good. Resolved two FIXMEs that were now possible to fix

Squash and merge 👍

PR #225, opened 2022-03-23
Fix #213 (Swipe ticket flow should match design)

### LINKED ISSUE #213: Swipe ticket flow should match design
In the ideal situation, the tapped ticket card should fly down to the bottom and change its content to a slider.
In the more realistic (for now) situation we have the same card design, but it appears from the bottom like a bottom modal sheet.

marfavi:
resolves #213 

Updated design of the "swipe to use ticket" view.
Now is a card that looks like a ticket/coffee card.
Furthermore, added hero animation between the two cards.

Replaced the slider widget.
Now uses a modified version of imtoori/flutter-slide-to-act
(ironed out some kinks and stripped down to meet our needs).

Tidied up


fremartini:
what was wrong with the previous slider? Did it not support all required functionality? Also is the slider not too large compared to the spec?

actual:
image

expected:
image

marfavi:
> what was wrong with the previous slider? Did it not support all required functionality?

The old slider determined its width based on the device screen size (its width would always be 90% of the screen width). This made the padding different based on what device you are using, see screenshot

image

The replacement widget uses up all the horizontal space it can use. It's far from perfect, but it serves as a good starting point (specifically, the code to determine the slider's width is really clever IMO).

> Also is the slider not too large compared to the spec?

The slider height in the spec was chosen because the height of the swipe card would then be the same height as the ticket/coffeecard card. (the idea was to minimize problems during the hero animation between the two cards).

Obviously, we want to decide the height based on how easy/comfortable it is to hit instead. Flutter doesn't seem to care about the height difference between the two cards while doing the hero animation anyway.

---

I just checked the height of the slider on a small device and determined it could be less tall on smaller devices. So now it's less tall on small devices, see screenshot

image

fremartini:
IMO the slider is still way too big. I would expect it to be as slim as the one in Figma, now it is nearly double the height

marfavi:
How about this?

image

fremartini:
> How about this? @fremartini
> 
> image

yes this looks good! :)

fremartini (approved):

PR #235, opened 2022-03-30
Fix user send to home-page after purchase

TTA777:
After a purchase, the user is now navigated back to the home_page if it was successful. Otherwise, they stay on the product page

### DIFF
--- a/lib/widgets/components/purchase/purchase_overlay.dart
+++ b/lib/widgets/components/purchase/purchase_overlay.dart
@@ -10,15 +10,13 @@ import 'package:coffeecard/widgets/components/purchase/purchase_process.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 
-class PurchaseOverlay {
-  final BuildContext _context;
-
-  Future<Payment?> show(
-    InternalPaymentType paymentType,
-    Product product,
-  ) async {
-    return showDialog<Payment>(
-      context: _context,
+Future<Payment?> showPurchaseOverlay({
+  required BuildContext context,
+  required Product product,
+  required InternalPaymentType paymentType,
+}) async =>
+    showDialog<Payment>(
+      context: context,
       barrierColor: AppColor.scrim,
       barrierDismissible: false,
       useRootNavigator: true,
@@ -55,11 +53,3 @@ class PurchaseOverlay {
         );
       },
     );
-  }
-
-  PurchaseOverlay.__create(this._context);
-
-  factory PurchaseOverlay.of(BuildContext context) {
-    return PurchaseOverlay.__create(context);
-  }
-}
--- a/lib/widgets/components/purchase/purchase_overlay.dart
+++ b/lib/widgets/components/purchase/purchase_overlay.dart
@@ -1,26 +1,23 @@
 import 'package:coffeecard/base/style/colors.dart';
-import 'package:coffeecard/cubits/environment/environment_cubit.dart';
 import 'package:coffeecard/cubits/purchase/purchase_cubit.dart';
 import 'package:coffeecard/cubits/receipt/receipt_cubit.dart';
 import 'package:coffeecard/cubits/tickets/tickets_cubit.dart';
-import 'package:coffeecard/models/receipts/receipt.dart';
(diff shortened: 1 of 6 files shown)


lib/widgets/components/purchase/purchase_overlay.dart
```diff
@@ -1,26 +1,23 @@
...
 class PurchaseOverlay {
   final BuildContext _context;
 
-  void hide() {
-    Navigator.of(_context, rootNavigator: true).pop();
-  }
-
-  void show(InternalPaymentType paymentType, Product product) {
-    showDialog(
+  Future<Payment?> show(
```
marfavi, line 16: I like the changes so far.

Consider if this class with a method should be refactored into a function `showPurchaseOverlay(context, paymentType, product)` instead. Currently it holds no state and I don't see a reason for it to stay as a class.

See Material functions showDialog, showModalBottomSheet, and our own function appDialog.

lib/widgets/components/purchase/purchase_overlay.dart
```diff
@@ -37,35 +34,22 @@ class PurchaseOverlay {
...
-                  hide(); //Removes this overlay
-                  // FIXME: Correct navigation
-                  Navigator.pop(context);
+                  final payment = state.payment;
+                  payment.productName =
+                      product.name; //TODO Receive this from backend
+                  Navigator.pop<Payment>(context, payment);
+
                   //TODO Consider if these calls should be moved elsewhere, e.g. inside the purchase cubit
                   final ticketCubit = sl.get<TicketsCubit>();
```
marfavi, line 43: These final lines of code should just be:

```dart
sl.get<TicketsCubit>().getTickets();
sl.get<ReceiptCubit>().fetchReceipts();
```

No need to await, as it doesn't do anything.

lib/widgets/components/tickets/buy_tickets_card.dart
```diff
@@ -24,15 +30,36 @@ class BuyTicketsCard extends StatelessWidget {
...
+      onTap: (context) async {
+        final payment = await showModalBottomSheet<Payment>(
           context: context,
           barrierColor: AppColor.scrim,
           backgroundColor: Colors.transparent,
           isDismissible: true,
           useRootNavigator: true,
           builder: (_) => BuyTicketBottomModalSheet(product: product),
         );
+        //ignore: use_build_context_synchronously
```
marfavi, line 42: Don't ignore linter errors

marfavi (changes requested):
Check the comments.

lib/widgets/components/purchase/purchase_overlay.dart
TTA777: In order to avoid a confusing situation in the future, if someone were to expand upon the method, I would definitely say that we need to await them. Even if it does not currently serve a purpose to do so. 

In #237 I did simplify the calls by not assigning the cubits to variables, as you suggest here. DO you want it as part of this PR, or are you fine with it being in the other one, considering I am moving where these calls happen anyway?

lib/widgets/components/purchase/purchase_overlay.dart
TTA777: Well, it could be refactored like that easily. I think the main argument for not doing so is that our over overlays work in this manner instead, i.e. the receipt and loading overlays. Mind you, for those two it is also to provide ourselves with the convenience of the hide method, so we always use the right navigator. That argument no longer applies to this class, so I would be fine with refactoring, if you still prefer that.

lib/widgets/components/purchase/purchase_overlay.dart
marfavi: Fine as a part of another PR

lib/widgets/components/purchase/purchase_overlay.dart
marfavi: Yes I would prefer refactoring into a function.

For all three overlays we want to use the root navigator, which is always accessible with `Navigator.of(context, rootNavigator: true)` and `showDialog(useRootNavigator: true, [other args...]);`. So I think that the other two classes also need to be refactored, but that can be looked at later.

marfavi (approved):
OK

PR #236, opened 2022-03-30
suggested changes to PR

marfavi:
splash pages split into loading and error pages
cleaned up some logic in splash router


TTA777 (approved):
lgtm, though I'll be honest that I don't quite agree with the linter as to the necessity of checking the widget is mounted, since the user will be trapped in the flow until the app helps the user out of there. Hence, why I chose to ignore the linter rule.

marfavi:
I see what you mean, but if the mounted check weren't there, the LoadingOverlay.of(context) call would throw an error anyway.

PR #240, opened 2022-04-01
Fix #239 (Navigating on login/logout fails)

### LINKED ISSUE #239: Navigating on login/logout fails
When attempting to log in, user gets stuck on loading overlay on login page

When logging out, user gets stuck on settings page

Tested on iOS and Android

marfavi:
Splash router needs to sit above the MaterialApp widget; otherwise, it gets removed from the navigation stack.

* Removed the empty EntryRouter widget.


lib/widgets/routers/splash_router.dart
```diff
@@ -1,31 +1,44 @@
...
+  void _authListener(BuildContext _, AuthenticationState state) {
+    _authStatus = state.status;
+    _maybeNavigate();
+  }
+
+  void _envListener(BuildContext _, EnvironmentState state) {
+    if (state is EnvironmentError) {
+      // FIXME: Handle error
+    }
+    _isTestEnvironment = state is EnvironmentLoaded && state.isTestEnvironment;
```
TTA777, line 31: This assignment is never null, so the later check for whether the environment is null, will always be false

lib/widgets/routers/splash_router.dart
```diff
@@ -1,31 +1,44 @@
...
+    _maybeNavigate();
+  }
 
-    if (authStatus.isUnknown || envState is! EnvironmentLoaded) return;
+  /// Navigates to the appropriate flow when both
+  /// _authStatus and _environment are loaded.
+  void _maybeNavigate() {
+    if (_authStatus.isUnknown || _isTestEnvironment == null) return;
```
TTA777, line 38: Due to the check on the environment always being false, as soon as the auth state is known, we will get redirected to the login screen. This happens almost instantly, since it just needs to read from local storage

TTA777 (changes requested):
First off, I'm unable to reproduce the error you are describing, so perhaps you could elaborate on what it is you are attempting to fix? The changes to the splash_routers logic broke it, so that will need to be fixed, if we want to pursue this PR any further. Perhaps just revert those, i.e. the methods called by the listeners

TTA777 (approved):

PR #241, opened 2022-04-03
Rewrite code for entry flows

marfavi:
## Design adjusted for entry (login, register, forgot passcode) flows
* Numpad contains forgot passcode action button
* Login passcode page contains a back button
* Splash screen now contains Analog logo
* Login pages now fully compatible with smaller devices (should never scroll or overflow on iPhone 5S-sized screens)

## Design implementations
* AppScaffold without title now adjusted; it's taller now, but it now contains both a back button as well as a environment warning button (if applicable).
* Removed biometric as an action on the numpad

## Business logic
* LoginBloc rewritten
  * now only responsible for passcode logic
  * All email validation and navigation related logic removed
  * LoginPageEmail now handles email verification logic
* Made less widgets dependent on BlocBuilder. now they instead rely on fields

## ✨Animations ✨
* splash screen logo animates to login screen
* numpad flies in from bottom
  * loosely related: the login email text field now receives focus after a delay to let page transitions finish first
* login passcode dots fade slightly
* page transitions between login email and passcode pages added
* Default page transition changed

## Other
* Forgot passcode
  * the text field now has an initial value of whatever email the user was trying to log in to
  * Added a field in EmailBody to make that possible
  * No longer exits the page on error
* abstract LoginPage widget replaced with LoginPageBase
* abstract RegisterPage widget replaced with RegisterPageBase
* RegisterFlow
  * Moved to widgets/pages/register folder
  * simplified; now wraps AppFlow
* LoginRouter removed
* LoginTitle removed (now a part of LoginPageBase)


lib/widgets/components/entry/login/login_email_text_field.dart
```diff
@@ -1,26 +1,55 @@
...
+
+class _LoginEmailTextFieldState extends State<LoginEmailTextField> {
+  final _controller = TextEditingController();
+  final _focusNode = FocusNode();
+
+  @override
+  void initState() {
+    // Focus the text field after animations have completed
+    () async {
+      await Future.delayed(widget.autoFocusDelay);
```
jonasanker, line 30: Can there arise a case where this will hang forever?

jonasanker (commented):
This has been a quick review. I had one comment else overall, I think the changes has been fine

lib/widgets/components/entry/login/login_email_text_field.dart
marfavi: No.

In the case of a really long set delay, the user can always tap the text field to focus the text field themselves

jonasanker: Okay. I'll approve

jonasanker (approved):

jonasanker (approved):

PR #247, opened 2022-04-26
Move cubits out of service locator

marfavi:
Fixes a bug with that redundantly runs the initial cubit method listed in home_page.dart and removes the need to reset cubits on logout


marfavi:
Quick writeup, will double check the code before taking it out of draft

jonasanker (approved):
Lgtm. Hadn't had the time to test it so trust you have 😉

PR #252, opened 2022-05-02
Fix #251 (User can change name to be empty)

### LINKED ISSUE #251: User's name can be empty or have leading/trailing whitespace
image

marfavi:
Resolves #251 

User could change name to contain leading/trailing whitespace or have an
empty name altogether.


lib/widgets/components/entry/register/name_body.dart
```diff
@@ -20,7 +20,7 @@ class NameBody extends StatefulWidget {
 
 class _NameBodyState extends State<NameBody> {
   final _controller = TextEditingController();
-  String get name => _controller.text;
+  String get name => _controller.text.trim();
 
   bool _showError = false;
   String? _error;
```
TTA777, line 26: Since we are trying to eliminate nullables in the codebase with the use of either, is there a reason that we are not using that here?

lib/widgets/components/entry/register/name_body.dart
```diff
@@ -37,22 +37,21 @@ class _NameBodyState extends State<NameBody> {
     super.initState();
   }
 
-  Future<void> _validateName(String name) async {
+  void _validateName() {
     setState(() => _error = name.isEmpty ? Strings.registerNameEmpty : null);
   }
 
   Future<void> _submit(BuildContext context) async {
+    _showError = true;
```
TTA777, line 45: Seems odd that we are deciding to show the error before we know whether there was one, or am I missing something?

TTA777 (changes requested):

lib/widgets/components/entry/register/name_body.dart
marfavi: This line of code was written before Either was implemented

lib/widgets/components/entry/register/name_body.dart
marfavi: Perhaps bad wording of variable. It means that it should display any error that will get encountered after this point. This is to prevent a validation error from displaying before the user submits.

marfavi: The `error` variable contains the actual error (and is `null` when no error is present)

lib/widgets/components/entry/register/name_body.dart
TTA777: I presumed as much, any reason we are not interested in updating it to using Either?

marfavi: The relevant widget takes in a `String?` for conditionally displaying the validation error. Using Either will therefore be more complex.

fremartini (approved):

TTA777 (approved):

PR #258, opened 2022-05-05
Reset nav bar app flows when tapping nav bar icon

### LINKED ISSUE #257: Navigating back to start of flow from nav bar
Tapping the navigation bar icon of the flow you are standing in should bring user back to start of flow.

Follows the convention set by other apps that have a navigation bar.

marfavi:
Fix #257

New navigation bar behaviour:

- Pressing the nav bar icon of a flow that is not currently active will bring the user to the page they were on before if any; otherwise default
  - (This was the old behaviour for all cases.)

- Pressing the nav bar icon of the flow that is currently active will bring the user back to the default page and scroll that page to the top.


fremartini (approved):
LGTM

PR #263, opened 2022-05-05
Replace leaderboard rank numbers with medal widget

### LINKED ISSUE #265: Use leaderboard rank medals for top three in leaderboard
Design spec:
image

marfavi:
Fix #265

To fit design spec

image


fremartini (commented):
LGTM, i would however prefer if the colors were moved to the lib/base/style/colors.dart file together with the rest of the colors. @marfavi what do you think?

fremartini (approved):

PR #279, opened 2022-05-09
Add variable flexibility of `ListEntry`

### LINKED ISSUE #280: SettingsListEntry not truncating text on right side and aligning incorrectly
example with normal length email:
image

example with long email:
image

marfavi:
Fix #280 

Fully fixes any `ListEntry` jank once and for all (hopefully).

# Motivation
Before, all list entries had the left side widget wrapped in a `Flexible` widget.
This caused a problem with SettingsListEntry, because here, the "variable"-sized widget (e.g. user email text) was on the right side.
The previous solution for that problem was to place the right side widget in a fixed size constrained box, but that fix still doesn't play well with variable length emails.
The proposed solution introduces `sideToExpand` inside of `ListEntry`. It tells the ListEntry which side should be displayed
in full, and which side should be wrapped in a `Flexible` widget (which allows that side to reduce in size based on remaining space).

## Misc
Used correct `SettingDescription` widget for the right side of each `SettingListEntry` that had text on their right side.

### DIFF
--- a/lib/widgets/components/list_entry.dart
+++ b/lib/widgets/components/list_entry.dart
@@ -3,7 +3,14 @@ import 'package:coffeecard/widgets/components/helpers/tappable.dart';
 import 'package:flutter/material.dart';
 import 'package:gap/gap.dart';
 
+/// Represents the left or right side of a [ListEntry] widget
+enum ListEntrySide { left, right }
+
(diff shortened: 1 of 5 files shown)


marfavi:
Maybe `sideToExpand` is a bit confusing, any thoughts on this name?

PR #287, opened 2022-05-11
display credits

fremartini:
closes #287 

* add credits page
* move api error formatting to extension method rather than Strings


marfavi (approved):
LGTM!

Issue #291, opened 2022-05-13
Pressing back button doesn't go back in current flow (recession)


marfavi:
Example: Going to Settings -> Opening hours and pressing back the button will not go back to Settings, but rather the previous flow (Tickets flow).

Issue #288, opened 2022-05-11
Programme page overflows


fremartini:
![Screenshot_20220511-222231.jpg](https://user-images.githubusercontent.com/39860858/167941217-a1441788-b33d-4232-9a2e-c19bb854dd14.jpg)

fremartini:
@marfavi this seems to be a problem caused by ListEntrySide as the text does not overflow but pushes the radio button outside the screen. I think a possible solution could be to swap the radiobutton and the text, what do you think? Otherwise we may need to introduce a ListEntrySide.None option that wouldn't expand any of the sides

marfavi:
> @marfavi this seems to be a problem caused by ListEntrySide as the text does not overflow but pushes the radio button outside the screen. I think a possible solution could be to swap the radiobutton and the text, what do you think? Otherwise we may need to introduce a ListEntrySide.None option that wouldn't expand any of the sides

Just need to change the side to expand. SettingListEntry has that set to left, when it should be right for the list entries here

marfavi:
Added two PRs that fix the issue because I came up with a better solution after the first PR.

Feel free to merge either one

Issue #261, opened 2022-05-05
Name in leaderboard entry can overflow


marfavi:
![image](https://user-images.githubusercontent.com/21163286/167020296-8976c2c4-a179-4afd-8159-9d554ff938d3.png)

PR #293, opened 2022-05-16
Add tests

### LINKED ISSUE #102: Discuss and agree on test strategies
We should discuss and agree on a test strategy

- What should be tested
- Type of testing
- Test coverage

fremartini:
closes #102 

i propose we at present time, start testing:
* All cubits (test if they emit correct states)
* All repositories (mock requests calls, return data object)

at a later point we may want to start widget testing, at present time i would argue that this kind of testing can be omitted. This would also mean when modifying above mentioned when implementing features, appropriate tests should also be made. I have created some tests to show how it could be done. Thoughts? @marfavi @jonasanker @TTA777


jonasanker:
I'm a big fan of this initiative! Kudos @fremartini 🙌

I think we need to find the sweet spot of testing where we test critical functionality also save the hassle of creating tests which are too sensitive for (non-breaking) changes.

I think cubits as starting point are good candidates, both state emitions and data transformations.

For repositories, these should also be tested but I think we have to reason about level, maybe mocking request and responses for repository methods where there are no object transformation, does not provide an interesting test case?

I'll think a bit more and reply back

fremartini:
> I think the UI is what breaks the most. So far we have merged in some branches that required a hotfix patch immediately.
> 
> Here's some recent UI bugs/recessions that could have been caught in a test (instead of requiring a fix branch):
> * #291
> * #288
> * #280
> * #267
> * #261
> * #251

Good point. I don't know how to properly test UI in Flutter yet. Could you push a proof-of-concept test to this branch? Otherwise i suggest we prioritize the logic for now

jonasanker:
UI tests are really cool but my experience is that they are also high effort. I think it could be fun to proof of concept them.

I'm not sure how well the library has been adapted to Flutter but I would suggest trying out the Selenium library.

fremartini:
I think we should make a choice here. While i think UI testing is important, i would suggest we start of only testing the cubits and repositories for now. Then create an issue concerning the UI testing and save that for later so testing doesn't stagnate. I'm basing this on the assumption that none of us have experience in testing UI. @jonasanker did you find a good way of testing the repositories? also what do you guys think? @marfavi @TTA777

jonasanker:
@fremartini Let me prototype something tomorrow

jonasanker:
I think it cool to prototype UI but agree it has less priority

marfavi:
> I think we should make a choice here. While i think UI testing is important, i would suggest we start of only testing the cubits and repositories for now. Then create an issue concerning the UI testing and save that for later so testing doesn't stagnate. I'm basing this on the assumption that none of us have experience in testing UI. @jonasanker did you find a good way of testing the repositories? also what do you guys think? @marfavi @TTA777

I agree with what you said. I created an issue for UI tests.

fremartini:
So whats the status on this one @marfavi ?

marfavi:
> So whats the status on this one @marfavi ?

Fixmes need to be addressed

fremartini:
> > So whats the status on this one @marfavi ?
> 
> Fixmes need to be addressed

Well the FIXME's dont belong in this PR and should be seperate issues. I'm converting them to TODO's instead

marfavi (approved):
OK

PR #298, opened 2022-05-19
convert existing FIXME's to TODO's

fremartini:
* convert all existing FIXME's to TODO's
* add entry in analysis_options to display them as info

'FIXME' is for things which are definitely broken, but where you want to not worry about it for the moment.
'TODO' is for useful features, optimizations or refactorings that might be worth doing in the future.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -23,3 +23,4 @@ analyzer:
     missing_return: error
     dead_code: info
     argument_type_not_assignable: warning
+    todo: info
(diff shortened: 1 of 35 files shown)


analysis_options.yaml
```diff
@@ -23,3 +23,4 @@ analyzer:
     missing_return: error
     dead_code: info
     argument_type_not_assignable: warning
+    todo: info
```
marfavi, line 26: We also talked about bumping FIXME to warning level errors, should we do that?

marfavi (approved):
LGTM
Check my commits and squash and merge if everything looks good to you.

PR #307, opened 2022-05-24
feat: save encoded passcode in secure storage

marfavi:
Move encoding responsibility out of account_repository;
Now all its methods expects already encoded passcodes.

Variable and method names for passcodes appropriately named `encodedPasscode` if they expect to be encoded.

Moved encode function to `utils/encode_passcode.dart`

---
Addresses

"TODO: Should probably have another return type in order to support
the intended registration flow?"
in `lib\data\repositories\shared\account_repository.dart`


jonasanker (approved):
LGTM

PR #309, opened 2022-05-26
Add fixed height for `ShopCard`s

marfavi:


marfavi:
Test on iOS devices

marfavi:
Using magic numbers will cut off text for people with large text scaling on their phones. We cannot use this solution because we want the app to look non-broken on as many devices as possible

PR #318, opened 2022-05-31
Open Privacy Policy in external browser

### LINKED ISSUE #165: Link to Privacy Policy in Settings page
A link to the app's privacy policy should be added to the settings page

jonasanker:
Closes #165

### DIFF
--- a/lib/base/strings.dart
+++ b/lib/base/strings.dart
@@ -299,6 +299,7 @@ abstract class Strings {
 
   // Errors
   static const error = 'Error';
+  static const cantLaunchUrl = 'The app can not launch the Url';
 
   static String unknownFilterCategory(Object category) {
     return 'Unknown filter category: $category';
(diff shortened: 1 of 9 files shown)


jonasanker:
I choose to open it using an external browser (instead of WebView) as the user will download a PDF and not read a webpage as such

marfavi (approved):
OK, I made `Privacy policy` sentence case and changed the `About Cafe Analog` section header to just say `About` (as intended).

marfavi (changes requested):
Use the `launchURL` method instead (that already launches an url in an external window).

lib/base/strings.dart
```diff
@@ -297,6 +299,7 @@ abstract class Strings {
 
   // Errors
   static const error = 'Error';
+  static const cantLaunchUrl = 'The app can not launch the Url';
```
marfavi, line 302: How about "Couldn't find an app to open $url"?

jonasanker: I had the same thought but the URL might not always be meaningful for a user e.g. MobilePay Deep links etc. Then what could make sense would probably be to log the errorful url

marfavi (approved):

PR #320, opened 2022-06-01
Release app (PATCH)

jonasanker:


marfavi (commented):
PR name needs a version tag `(MINOR)`

jonasanker:
> > @marfavi When no tag is added, the release will automatically be a `PATCH` release when I believe this should be :)
> 
> Alright. I still think we should explicitly always add the tag regardless to express intent of the scope of the release. What do you think?

I see your point. I actually don't think the plugin supports listening to the `(PATCH)` message. I have changed the auto-commit action to include the patch in the commit message

marfavi (approved):

PR #339, opened 2022-08-18
sign user out when confirming account deletion

### LINKED ISSUE #328: User should be logged out of app after request of account deletion
The user should be logged out of the app after he has requested account deletion so that he will experience no funky state trying to use the app with a token which is not valid.

fremartini:
closes #328


marfavi (approved):

fremartini:
> OK. User can still log back in to their account and then go to the confirmation email, but I think fixing that requires changes to token invalidation in the backend.

Yes, that would require extending the users state in the backend but i believe this is sufficient for now

PR #346, opened 2022-09-01
Update FAQ page

marfavi:
(misc) Update iOS build files

Replaced ExpansionTile with custom FAQ widget

Adjust padding sizes


marfavi:
> LGTM, i noticed that the animation when expanding the dropdown is no longer present which was included in ExpansionTile. Is this something we want?

No, I don't think it's important. The implementation used in ExpansionTile is clunky as well.

fremartini (approved):

PR #354, opened 2022-10-25
display user gravatar

### LINKED ISSUE #194: Use Gravatar to generate user picture
The following call can be used to generate a Gravatar from a users email

https://www.gravatar.com/avatar/${email}?d=identicon&s=50

fremartini:
closes #194


lib/widgets/components/gravatar_image.dart
```diff
@@ -0,0 +1,38 @@
...
+      imageBuilder: (context, imageProvider) => Container(
+        width: size,
+        height: size,
+        decoration: BoxDecoration(
+          shape: BoxShape.circle,
+          image: DecorationImage(image: imageProvider, fit: BoxFit.cover),
+        ),
+      ),
+      imageUrl:
+          'https://gravatar.com/avatar/${hashMD5(hash)}?s=100&d=identicon',
```
jonasanker, line 26: This might be an over-engineering suggestion, but to plan a bit for the future, then would it make sense to wrap this call in a Repository class as we do for other RestApi calls such that the UI do not care about this task

fremartini: Removed Gravatar in favor of colored background

lib/utils/encode_passcode.dart
```diff
@@ -1,9 +1,13 @@
...
-import 'package:crypto/crypto.dart' show sha256;
+import 'package:crypto/crypto.dart' show md5, sha256;
 
 String encodePasscode(String passcode) {
   final bytes = utf8.encode(passcode);
   final passcodeHash = sha256.convert(bytes);
   return base64Encode(passcodeHash.bytes);
 }
+
+String hashMD5(String input) {
```
TTA777, line 11: Consider renaming this file, now that it no longer only encodes the password

lib/widgets/components/user_icon.dart
```diff
@@ -0,0 +1,44 @@
+import 'package:coffeecard/widgets/components/images/coffee_image.dart';
```
TTA777, line 1: Since we now have two components related to the user, should they not be in a folder for themselves?

TTA777 (changes requested):

lib/utils/encode_passcode.dart
TTA777: Since you ended up not making use of gravatars, consider simply deleting the unused function as well

lib/widgets/pages/settings/your_profile_page.dart
```diff
@@ -57,7 +58,9 @@ class _EditProfile extends StatelessWidget {
     return Column(
       children: [
         const Gap(24),
-        const CircleAvatar(radius: 54),
+        UserIcon.large(
+          id: user.id,
```
marfavi, line 62: Can be one line

TTA777 (approved):
lgtm

PR #359, opened 2022-11-08
Fix/frem/reactivation authentication

### LINKED ISSUE #358: Verify user token on app start and kick user out after 401
Actual behaviour:
* User can access the main flow with an invalid user token. (all subsequent requests would return 401)
* User is not kicked out of the main flow after a request returns 401.

Expected behaviour:
* On app start, the saved user token should be verified OR a new one should be generated with the help of saved user credentials
* The user should be kicked out of the main flow after a request returns 401.

fremartini:
closes #358

### DIFF
--- a/lib/utils/mutex.dart
+++ b/lib/utils/mutex.dart
@@ -11,7 +11,10 @@ class Mutex {
   }
 
   void unlock() {
-    _completer.complete();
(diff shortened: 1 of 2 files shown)


marfavi:
I'll verify that it works in the weekend

marfavi (approved):
OK

PR #361, opened 2022-11-08
Placeholder for statistics page when user is unranked

### LINKED ISSUE #357: Display placeholder string when user rank is '0th'
Currently, when a user does not have a rank the corresponding card displays '0th' place.

The user rank should display '-' when the user has no rank.
When the user is not displayed on the leaderboard and has a rank, it should display their current rank in the leaderboard section

duckth:
closes #357 

When the user has ro rank, instead of displaying "0th" on the statistics page, a dash '-' will now be displayed instead.


marfavi (approved):
good

fremartini (approved):

PR #363, opened 2022-11-15
Use Executor for repository calls; Make TicketsSection respond to error states

### LINKED ISSUE #362: Handle errors occurring while using a ticket
While the TicketsCubit does emit error states, the TicketsSections listening on it does not react to these. The result is if an error occurs while swiping a ticket, the user will be trapped on a loading spinner. 

How to reproduce:
1. Open the app to the homescreen
2. Turn off internet on the device
3. Swipe a ticket
4. Be trapped on a loading spinner since the error state is not handled

marfavi:
Use new Executor utility class to handle API calls 

fixes #362 by ensuring that `TicketRepository.useTicket` always returns an Either, and that the `TicketsSection` responds to TicketsCubit error states.


marfavi:
From [chopper's documentation (bottom of page)](https://hadrien-lejard.gitbook.io/chopper/getting-started), it seems like this PR uses the intended solution. 

> Handling I/O and other exceptions should be done by surrounding requests with try-catch blocks.

marfavi:
Still some FIXMEs that need be to addressed before this branch is ready

fremartini (approved):

PR #364, opened 2022-11-16
POC, for shared utility handling network errors

TTA777:
Here is a POC of what I was trying to suggest yesterday. I've only updated one of the repositories to make use of it, though almost all repos should be similar small changes. 

Personally, I prefer this approach since we do the same error handling anyway, in all the repos. What I think is most debatable is whether the utility should be a static method, or if it should be used on an instance and then inject the instance where needed. I lean to making it a static method, but welcome arguments to the contrary.


marfavi:
Check #365

marfavi (approved):

PR #365, opened 2022-11-17
Handle api calls with Executor

marfavi:


marfavi:
Executor handles errors, and in cases of successful api return, transforms the response type. 

Executor is injected as an instance into repositories.

Logger is injected as an instance into Executor. All api failures are now logged properly.

Logger is no longer directly injected into any repository.

lib/data/repositories/v2/app_config_repository.dart
```diff
@@ -1,49 +1,34 @@
...
-    final Response<AppConfig> response;
-    try {
-      response = await _api.apiV2AppconfigGet();
-    } catch (e) {
-      return Left(ClientNetworkError());
+  Environment _onSuccessfulRequest(AppConfig dto) {
+    switch (environmentTypeFromJson(dto.environmentType as String)) {
+      case EnvironmentType.production:
+        return Environment.production;
+      case EnvironmentType.localdevelopment:
```
TTA777, line 20: Please add a comment that the fall through is intended

lib/data/repositories/utils/executor.dart
```diff
@@ -1,25 +1,30 @@
...
-    Future<Response<T>> Function() f,
-    Logger logger,
-  ) async {
-    final Response<T> response;
+  const Executor(this.logger);
+  final Logger logger;
+
+  /// Executes a network request and handles errors, including IO errors.
+  /// Returns either a [RequestError] or the transformed success type.
+  Future<Either<RequestError, Ret>> execute<Dto, Ret>(
```
TTA777, line 13: Can we call the returnType something else than Ret? Doesn't indicate it's usage very well

lib/data/repositories/utils/executor.dart
```diff
@@ -1,25 +1,30 @@
...
+    Future<Response<Dto>> Function() request, {
+    required Ret Function(Dto dto) transformer,
```
TTA777, line 15: Why are we giving the parameters in two different ways?

TTA777 (changes requested):
Minor comments, overall looks good

lib/data/repositories/utils/executor.dart
marfavi: Called it RightType. Slightly changed doc comment.

lib/data/repositories/utils/executor.dart
marfavi: both are now positional

jonasanker (changes requested):

lib/data/repositories/utils/executor.dart
```diff
@@ -1,25 +1,32 @@
...
+        return Right(transformer(response.body as Dto));
+      } else {
+        logger.e('API Error: (${response.statusCode}) ${response.error}');
+        return Left(RequestError.fromResponse(response));
       }
-    } catch (e) {
-      return Left(ClientNetworkError());
+    } on Exception catch (e) {
+      logger.e('HTTP Client Error: $e');
+      return Left(RequestError.clientNetworkError());
```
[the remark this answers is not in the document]
marfavi: caught specific errors related to http client

lib/data/repositories/utils/request_types.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:chopper/chopper.dart';
+import 'package:coffeecard/base/strings.dart';
+
+/// Empty success class.
+class RequestSuccess {}
+
+class RequestError {
+  RequestError(this.message, this.code);
+  final String message;
+  final int code;
```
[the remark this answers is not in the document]
marfavi: Renamed to RequestFailure with only one field: message. Made RequestHttpFailure that also includes status code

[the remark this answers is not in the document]
marfavi: Now that field `code` belongs to RequestHttpFailure, it should no longer be ambigious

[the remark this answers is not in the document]
marfavi: Reasoning behind renaming RequestError -> RequestFailure is to stick to [dart's error naming conventions](https://groups.google.com/a/dartlang.org/g/misc/c/lx9CXiV3o30/m/s5l_PwpHUGAJ).

Types named Error or Exceptions are meant to be thrown. We don't throw RequestFailure

PR #367, opened 2022-11-20
Use variable font; Build text styles with `TextStyleBuilder`

marfavi:
Default font for app changed to Roboto Flex. Beforehand, the app would use Roboto on Android and SF UI on iOS. We wanted to use SF Pro Display/Text on iOS for its different optical sizes, but this has proven impossible to achieve on Material-based Flutter apps. Instead, we will use the variable font Roboto Flex to achieve fluid optical sizing on both platforms.

Due to the new font's flexibility (heh) we create presets of the font for heading and body text, just like how SF Pro has a Display and Text variant. As a result, app font now matches the design spec pretty closely.

A new `TextStyleBuilder` class will handle setting variable font axes depending on if the text style is for heading or body text.

misc: AppTextStyle.medium renamed to AppTextStyle.botomNavBarLabel

misc: SectionTitle now uses AppTextStyle.sectionTitle text style.


lib/widgets/components/login/login_numpad.dart
```diff
@@ -87,7 +87,7 @@ class _NumpadState extends State<Numpad> with SingleTickerProviderStateMixin {
                         verticalAlignment: TableCellVerticalAlignment.fill,
                         child: _NumpadButton(
                           onPressed: widget.forgotPasscodeAction,
-                          child: const Text(
+                          child: Text(
                             'Forgot?',
```
fremartini, line 91: move to strings

fremartini (approved):

fremartini (approved):

fremartini (approved):

PR #373, opened 2022-11-23
bump & enforce iOS deployment target

marfavi:
iOS build step fails on PR #372 on warning:
```
warning: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 15.2.99. (in target 'FMDB' from project 'Pods')
```
There have been other instances of this error occurring locally.

This fix hopefully fixes the problems.


jonasanker:
@marfavi Before merging, please comment on why we want to deprecate iOS 11 versions. It's a breaking change so we want to be transparent about why

PR #374, opened 2022-11-23
Release App (MINOR)

jonasanker:
- Re-added paymentTypeToJson so dto can be deserialized (#371)
- Fix "in the future" when diff is very small (#372)


marfavi:
This should be a PATCH level release

PR #392, opened 2022-12-20
fix map parse error

fremartini:


duckth (approved):
Looks good to me, I've also tested that it works on my phone 👍

marfavi (approved):

PR #400, opened 2023-02-21
Add upgrade alert on login and ticket page

fredpetersen:


pubspec.yaml
```diff
@@ -80,6 +80,7 @@ dependencies:
 
   # functional programming thingies
   dartz: 0.10.1
+  upgrader: ^4.11.1
```
fremartini, line 83: this should probably have it own category and not fall under 'functional programming thingies'

fremartini (approved):

PR #402, opened 2023-03-04
Fix uncaught error in ServerException

marfavi:
Use Failures instead of exceptions where possible

Make message in ServerFailure.fromResponse nullable; catch FormatException (happens when empty string gets parsed)

Remove interface fo OpeningHoursRemoteDataSource, since its implementation was in the same file

Improve logic that generates opening hours from list of shifts; now all days that don't have shifts are closed (no assumptions are made about what days are closed)


fremartini:
@marfavi @TTA777 
https://github.com/AnalogIO/coffeecard_app/pull/404

fremartini:
closed by #404

PR #403, opened 2023-03-04
bump packages, dart and flutter version

fremartini:
We now have several
"Don't use 'BuildContext's across async gaps. Try rewriting the code to not reference the 'BuildContext'." 
warnings in the code, how do we address these?


fredpetersen (approved):
✅

PR #404, opened 2023-03-04
revision/fix/server exception uncaught exceptionrevision

fremartini:
* Refactor the usage of Executor in the data access layer.
* Removed the mapping functionality from the executor as this proved impossible to test with the current mocking framework.
* Handle case where jsonString['message'] was null in ServerException
* Removed RequestSuccess and replaced with void


fremartini:
@marfavi

lib/widgets/components/forms/forgot_passcode/forgot_passcode_form.dart
```diff
@@ -57,11 +57,11 @@ class ForgotPasscodeForm extends StatelessWidget {
     var body = Strings.forgotPasscodeSentRequestTo(email);
```
marfavi, line 57: Simplify
```suggestion
    final title = either.fold(
      (error) => Strings.forgotPasscodeError,
      (_) => Strings.forgotPasscodeLinkSent,
    );
    final body = either.fold(
      (error) => error.reason,
      (_) => Strings.forgotPasscodeSentRequestTo(email),
    );
```

test/core/network/executor_test.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  test('should return [ServerFailure] if call throws [Exception]', () async {
+    // arrange
+    final tException = Exception('some error');
+
+    // act
+    final actual = await executor(() async => throw tException);
+
+    // assert
+    expect(actual, const Left(ServerFailure('connection refused')));
```
marfavi, line 51: Magic string; The reason "connection refused" is dependent on the value written in the Executor class. Hopefully, the suggested changes in my review will also address this as a side effect

test/core/network/executor_test.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  test('should return [ServerFailure] if api call fails', () async {
+    // arrange
+    final tResponse = Response.fromStatusCode(500);
+
+    // act
+    final actual = await executor(() async => tResponse);
+
+    // assert
+    expect(actual, const Left(ServerFailure('')));
```
marfavi, line 29: Check the type instead of checking against the concrete value `Left(ServerFailure(''))`

test/data/repositories/v2/ticket_repository/ticket_repository_test.dart
```diff
@@ -8,7 +7,7 @@ import 'package:logger/logger.dart';
 import 'package:mockito/annotations.dart';
 import 'package:mockito/mockito.dart';
 
-import '../../responses.dart';
+import '../../../../response.dart';
```
marfavi, line 10: Avoid relative imports

test/features/opening_hours/data/datasources/opening_hours_remote_data_source_test.dart
```diff
@@ -1,85 +1,54 @@
...
   setUp(() {
     api = MockShiftplanningApi();
-    dataSource = OpeningHoursRemoteDataSourceImpl(api: api);
+    executor = MockExecutor();
+    dataSource = OpeningHoursRemoteDataSourceImpl(api: api, executor: executor);
   });
 
   group('isOpen', () {
-    test('should throw [ServerException] if api call fails', () async {
+    test('should call executor', () async {
```
marfavi, line 24: To check if the executor was called, use

```dart
      // check if executor was called with the correct function
      verify(executor<List<OpeningHoursDTO>>(any)).called(1);
```

test/features/opening_hours/data/repositories/opening_hours_repository_impl_test.dart
```diff
@@ -70,10 +47,15 @@ void main() {
         const Right(
           OpeningHours(
             allOpeningHours: {
+              1: 'Closed',
```
marfavi, line 50: Use `DateTime.monday` etc instead of ints

```suggestion
               DateTime.monday: 'Closed',
```

test/features/opening_hours/domain/usecases/check_open_status_test.dart
```diff
@@ -7,25 +7,25 @@ import 'package:mockito/mockito.dart';
...
-    when(repository.getIsOpen()).thenAnswer((_) async => const Right(true));
+    when(dataSource.isOpen()).thenAnswer((_) async => const Right(true));
 
     // act
     await getIsOpen(NoParams());
 
     // assert
-    verify(repository.getIsOpen());
-    verifyNoMoreInteractions(repository);
+    verify(dataSource.isOpen());
```
marfavi, line 28: add `.called(1)` to check for only a single call

```suggestion
    verify(dataSource.isOpen()).called(1);
```

lib/data/repositories/v2/leaderboard_repository.dart
```diff
@@ -27,21 +27,24 @@ class LeaderboardRepository {
...
+  Future<Either<ServerFailure, List<LeaderboardUser>>> getLeaderboard(
     LeaderboardFilter category,
   ) async {
-    return executor.execute(
+    final result = await executor(
       () => apiV2.apiV2LeaderboardTopGet(preset: category.label, top: 10),
-      (dtoList) => dtoList.map(LeaderboardUser.fromDTO).toList(),
     );
+
+    return result
```
marfavi, line 37: Use map instead of bind

lib/data/repositories/v2/leaderboard_repository.dart
```diff
@@ -27,21 +27,24 @@ class LeaderboardRepository {
...
+  Future<Either<ServerFailure, LeaderboardUser>> getLeaderboardUser(
     LeaderboardFilter category,
   ) async {
-    return executor.execute(
+    final result = await executor(
       () => apiV2.apiV2LeaderboardGet(preset: category.label),
-      LeaderboardUser.fromDTO,
     );
+
+    return result.bind((result) => Right(LeaderboardUser.fromDTO(result)));
```
marfavi, line 48: Use map instead of bind
```suggestion
    return result.bind(LeaderboardUser.fromDTO);
```

lib/data/repositories/v2/purchase_repository.dart
```diff
@@ -16,28 +16,30 @@ class PurchaseRepository {
...
       () => apiV2.apiV2PurchasesPost(
         body: InitiatePurchaseRequest(
           productId: productId,
           paymentType: paymentTypeToJson(paymentType),
         ),
       ),
-      (dto) => InitiatePurchase.fromDto(dto),
     );
+
+    return result.bind((result) => Right(InitiatePurchase.fromDto(result)));
```
marfavi, line 32: Use map instead of bind
```suggestion
    return result.map(InitiatePurchase.fromDto);
```

lib/data/repositories/v2/purchase_repository.dart
```diff
@@ -16,28 +16,30 @@ class PurchaseRepository {
...
+  Future<Either<ServerFailure, SinglePurchase>> getPurchase(
     int purchaseId,
   ) async {
-    return executor.execute(
+    final result = await executor(
       () => apiV2.apiV2PurchasesIdGet(id: purchaseId),
-      (dto) => SinglePurchase.fromDto(dto),
     );
+
+    return result.bind((result) => Right(SinglePurchase.fromDto(result)));
```
marfavi, line 43:
```suggestion
    return result.map(SinglePurchase.fromDto);
```

test/features/opening_hours/domain/usecases/check_open_status_test.dart
fremartini: verify(dataSource.isOpen()) and verify(dataSource.isOpen()).called(1) is identical i believe. The .called(1) is implicit

test/data/repositories/v2/ticket_repository/ticket_repository_test.dart
fremartini: not possible here.

test/core/network/executor_test.dart
fremartini: no, i want to differentiate between the exception block and the unsuccessful block. 

the empty string is used because 'Returns the response body as a String provided the network call was successful, else this will be null' in ServerFailure.fromResponse:

 on Exception {
      return ServerFailure(response.bodyString);
    }

this is a limitation of the way we create dummy requests

test/core/network/executor_test.dart
fremartini: i would argue this string should not go in Strings.dart because this is not displayed in the UI and is perfectly fine to test for

lib/core/network/executor.dart
```diff
@@ -0,0 +1,29 @@
...
+
+      if (!response.isSuccessful) {
+        logger.e(response.toString());
+        return Left(ServerFailure.fromResponse(response));
+      }
+
+      return Right(response.body as Result);
+    } on Exception {
+      // could not connect to backend for whatever reason
+      return const Left(ServerFailure('connection refused'));
```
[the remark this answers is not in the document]
fremartini: i would argue this string should not go in Strings.dart because this is not displayed in the UI.

lib/core/network/executor.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:chopper/chopper.dart' show Response;
+import 'package:coffeecard/core/errors/failures.dart';
+import 'package:dartz/dartz.dart';
+import 'package:logger/logger.dart';
+
+class Executor {
```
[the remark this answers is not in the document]
fremartini: you can look at where and how its used, no need to maintain documentation. Otherwise the class could be renamed to NetworkExecutor or RequestExecutor. I personally this its perfectly clear what it does when looking at the implementation

lib/core/network/executor.dart
```diff
@@ -0,0 +1,29 @@
...
+  final Logger logger;
+
+  const Executor(this.logger);
+
+  /// Executes a network request.
+  Future<Either<ServerFailure, Result>> call<Result>(
```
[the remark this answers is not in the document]
fremartini: i think this adds excessive function calls that are not needed, and bloats any calling function. The class already contains only a single function that only does one thing. Otherwise the class could be renamed to NetworkExecutor or RequestExecutor.

lib/features/opening_hours/domain/repositories/opening_hours_repository.dart
```diff
@@ -3,6 +3,5 @@ import 'package:coffeecard/features/opening_hours/domain/entities/opening_hours.
 import 'package:dartz/dartz.dart';
 
 abstract class OpeningHoursRepository {
```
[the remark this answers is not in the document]
fremartini: i think we should talk about this in person

lib/features/opening_hours/data/datasources/opening_hours_remote_data_source.dart
```diff
@@ -1,44 +1,42 @@
-import 'package:coffeecard/core/errors/exceptions.dart';
+import 'package:coffeecard/core/errors/failures.dart';
+import 'package:coffeecard/core/network/executor.dart';
 import 'package:coffeecard/generated/api/shiftplanning_api.swagger.dart';
+import 'package:dartz/dartz.dart';
 
 abstract class OpeningHoursRemoteDataSource {
```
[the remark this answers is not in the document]
fremartini: i think we should talk about this in person

test/response.dart
```diff
@@ -1,8 +1,11 @@
 import 'package:chopper/chopper.dart' as chopper;
 import 'package:http/http.dart' as http;
 
-class ResponseFactory {
-  static chopper.Response<T> fromStatusCode<T>(int statusCode, {T? body}) {
+class Response {
```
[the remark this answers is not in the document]
fremartini: i think we should talk about this in person

lib/core/errors/failures.dart
```diff
@@ -12,6 +15,19 @@ abstract class Failure extends Equatable {
 // General failures
 class ServerFailure extends Failure {
```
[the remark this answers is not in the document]
fremartini: i think we should talk about this in person

fremartini:
@marfavi

lib/core/errors/failures.dart
```diff
@@ -9,9 +12,21 @@ abstract class Failure extends Equatable {
...
   const ServerFailure(super.reason);
+
+  factory ServerFailure.fromResponse(Response response) {
+    try {
+      final jsonString =
+          json.decode(response.bodyString) as Map<String, dynamic>;
+
+      final message = jsonString['message'] as String?;
+
+      return ServerFailure(message ?? response.bodyString);
```
TTA777, line 25: At our last meeting, we discussed changing this, so we wouldn't show the bodyString to the user if the message was null. Instead, we would show some default message like "An unknown error occurred", and then log the bodyString to firebase, so we could actually act on whatever error occurred. I think that is something we should go ahead and do, so that we get actionable data for ourselves, and the user does not see bogus info.

lib/data/repositories/shared/account_repository.dart
```diff
@@ -20,13 +20,13 @@ class AccountRepository {
   final CoffeecardApiV2 apiV2;
   final Executor executor;
 
-  Future<Either<RequestFailure, RequestSuccess>> register(
+  Future<Either<ServerFailure, void>> register(
```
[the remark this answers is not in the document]
TTA777: I would agree with marfavi here, that a class to signalling a success would be preferable to having void returned

lib/widgets/components/forms/forgot_passcode/forgot_passcode_form.dart
```diff
@@ -53,15 +53,13 @@ class ForgotPasscodeForm extends StatelessWidget {
...
-
-    either.fold(
-      (l) {
-        title = Strings.forgotPasscodeError;
-        body = l.message;
-      },
-      (r) => null,
+    final title = either.fold(
+      (_) => Strings.forgotPasscodeError,
+      (_) => Strings.forgotPasscodeLinkSent,
```
TTA777, line 58: For the sake of readability, we might want to use a name for the variables, instead of _, even though we are not using the values. Right now it is not apparent which of the cases is the error and success, unless you memorized the syntax of fold

TTA777 (changes requested):
One thing I want addressed, and a few minor comments

test/features/opening_hours/domain/usecases/check_open_status_test.dart
marfavi: (Counter-intuitively) it doesn't work like that. Try verifying by duplicating the call to `getIsOpen` above; the test still passes. Only when `.called(1)` is added to the verify function is the test invalidated by two calls to `getIsOpen`

marfavi (approved):
LGTM

test/features/opening_hours/domain/usecases/check_open_status_test.dart
marfavi: For a later conversation

PR #406, opened 2023-03-18
Rollback of packages to support our openapi spec.

TTA777:
Minor code changes to accommodate the rollback of packages


marfavi:
Alternative solution in the works - PR on hold in the meantime

PR #415, opened 2023-03-28
Fixed the authenticator not returning the updated request

TTA777:
Fixes an error in the reactivation_authenticator where it was not returning the value resulted from either.fold


fremartini (approved):

fremartini:
@TTA777 LGTM. I pushed a new commit moving the logging to a separate method to avoid duplication. Was there a reason bodyString was only logged in one branch?

fremartini:
> @fremartini My reasoning was that as long as the request is successful, I'm not really interested in what the body is, and I thought it was cluttering up the terminal. It might be silly to remove it, I'm not sure, but that was my reasoning

allright. I propose we merge it in then

PR #418, opened 2023-04-05
Removed http_utils

### LINKED ISSUE #387: remove http_utils.dart when backend maps errors correctly
http_utils.dart

### LINKED ISSUE #417: Login attempt using wrong creds stalls app
Regression in latest develop build. ~Might be related to #415~

Reproduce: Login using wrong credentials (wrong email/passcode combination) -- see video

https://user-images.githubusercontent.com/21163286/230072444-d242700f-7bee-45e2-93ec-5fe77ec46224.mp4

TTA777:
Closes #417 
Closes #387 

Its functionality has already been replaced by ServerFailure, and is now causing an issue where we are accidentaly calling it with just a string, which then triggers an exception


fremartini:
This issue was filed because when the user attempts to log in with an incorrect passcode, they would see the json object {'error': 'message'} so we would have to decode it. If the linked closed issues)prs have solved this or if the openapi spec has been updated this issue can be closed. Can someone verify this?

TTA777:
The OpenApi spec has not been updated yet, however all requests go through the executor now. If a request fails in the executor, we manually decode it in ServerFailure.fromResponse and return the ServerFailure, so the loginCubit would never get the raw json in the new implementation. 

Whether we choose to close the issue, or update it, so it pertains to Serverfailure instead, is fine with me.

fremartini (approved):

PR #421, opened 2023-04-14
Update codecov.yml

fremartini:


marfavi:
Does the coverage output from `make cov` produce the correct coverage files? Otherwise, we should consider changing the Makefile as well.

fremartini:
@marfavi should be the case now but i cannot verify as i don't have lcov installed. Can you verify?

marfavi (dismissed):

PR #430, opened 2023-04-23
HOTFIX: Fix usage of `sed` command on MacOS

marfavi:
Prevented the pipeline from building iOS app for production


TTA777 (approved):
lgtm

PR #431, opened 2023-04-23
HOTFIX: Fix usage of `sed` command on MacOS

marfavi:
Fixed a bug in the pipeline code that prevented the pipeline from building a production iOS app


TTA777 (approved):
lgtm

PR #432, opened 2023-04-23
Trigger release workflow on workflow change

marfavi:
Included release.yml to the list of paths that should trigger a release

https://github.com/AnalogIO/coffeecard_app/commit/7702fde49fc514fe1da8c05303bcd7af1726bea5 didn't trigger a release workflow (which we need right now), but turns out the list of paths to trigger the workflow might have been incorrect.


TTA777 (approved):
lgtm

PR #443, opened 2023-05-03
Refactor and test Leaderboard

fremartini:


lib/features/leaderboard/domain/usecases/get_leaderboard.dart
```diff
@@ -0,0 +1,70 @@
...
+    implements UseCase<List<LeaderboardUser>, LeaderboardFilter> {
+  final LeaderboardRemoteDataSource remoteDataSource;
+
+  GetLeaderboard({required this.remoteDataSource});
+
+  @override
+  Future<Either<Failure, List<LeaderboardUser>>> call(
+    LeaderboardFilter filter,
+  ) async {
+    final userleaderboard = await remoteDataSource.getLeaderboardUser(filter);
```
marfavi, line 18: Better naming could benefit this code
```suggestion
    final user = await remoteDataSource.getLeaderboardUser(filter);
```
```suggestion
    final leaderboardUser = await remoteDataSource.getLeaderboardUser(filter);
```

marfavi (changes requested):

lib/features/leaderboard/data/datasources/leaderboard_remote_data_source.dart
```diff
@@ -18,32 +20,31 @@ extension _FilterCategoryToPresetString on LeaderboardFilter {
...
     required this.apiV2,
     required this.executor,
   });
 
-  final CoffeecardApiV2 apiV2;
-  final NetworkRequestExecutor executor;
-
   Future<Either<NetworkFailure, List<LeaderboardUser>>> getLeaderboard(
     LeaderboardFilter category,
+    int top,
```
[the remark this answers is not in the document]
fremartini: i disagree. The data access layer does not need to know about business logic. The data layer wraps the backend that exposes the 'top' attribute which should be controlled by a use case or a repository

lib/features/leaderboard/data/models/leaderboard_user_model.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:coffeecard/features/leaderboard/domain/entities/leaderboard_user.dart';
+import 'package:coffeecard/generated/api/coffeecard_api_v2.models.swagger.dart';
+
+class LeaderboardUserModel extends LeaderboardUser {
```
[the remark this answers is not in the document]
fremartini: Yes this exists to seperate the data related model from the domain model. I agree that the domain model may not need to know if it should be highligted or not; to me that sounds like a UI related property. I'll have a look at this

test/core/network/network_request_executor_test.dart
```diff
@@ -37,32 +37,32 @@ void main() {
 
   test('should return [ServerFailure] if api call fails', () async {
     // arrange
-    final tResponse = responseFromStatusCode(500, body: '');
+    final testResponse = responseFromStatusCode(500, body: '');
```
fremartini, line 40: i thought we agreed on some strategy to name these kinds of variables? I think the most important aspect is to remain consistent. Everywhere else i have used tResponse, tSomething as in 'the response', 'the something'

marfavi: I agree we should be consistent. In 727297093cab26834a29c9cfcb3d637c732f48b5, I changed all occurrences of `t[Name]` to `test[Name]`. We should generally strive to avoid abbreviations in code and this seemed like a quick fix.

test/features/leaderboard/data/datasources/leaderboard_remote_data_source_test.dart
```diff
@@ -0,0 +1,113 @@
...
+  late MockNetworkRequestExecutor executor;
+  late LeaderboardRemoteDataSource dataSource;
+
+  setUp(() {
+    apiV2 = MockCoffeecardApiV2();
+    executor = MockNetworkRequestExecutor();
+    dataSource = LeaderboardRemoteDataSource(apiV2: apiV2, executor: executor);
+  });
+
+  const tErrorMessage = 'some error';
```
marfavi, line 26:
```suggestion
  const testErrorMessage = 'some error';
```

marfavi (changes requested):
Simple request on naming of variable. Rest looks good

lib/features/leaderboard/data/models/leaderboard_user_model.dart
```diff
@@ -0,0 +1,22 @@
...
+    required super.highlight,
+  });
+
+  factory LeaderboardUserModel.fromDTO(LeaderboardEntry dto) {
+    return LeaderboardUserModel(
+      id: dto.id!,
+      rank: dto.rank!,
+      score: dto.score!,
+      name: dto.name!,
+      highlight: false,
```
marfavi, line 19: A test fails because this value is `false`.

marfavi (approved):
LGTM

PR #452, opened 2023-05-13
Migrate to Dart 3 and Flutter 3.10.0

marfavi:
Closes #451

PR #457, opened 2023-05-14
migrate to fpdart

### LINKED ISSUE #445: Move away from `dartz` package
I heard there is a better, well-documented functional programming package for Dart named [fpdart](https://pub.dev/packages/fpdart). It specifically compares itself to dartz and critiques the package for having no documentation, something our team agrees with. It should be researched if fpdart covers our requirements, and if so, remove the dartz package with the fpdart package.

fremartini:
closes #445


marfavi:
I think a drop-in replacement is insufficient. fpdart should have a few tricks to improve our code quality, like the Do notation in the latest release.

I suggest changing this new pull request to target the branch migrate/dart3-and-fpdart (#452) instead, so we can explore the features of fpdart and apply any code improvements before we merge to develop

marfavi (changes requested):
Read my other comment

fremartini:
> I think a drop-in replacement is insufficient. fpdart should have a few tricks to improve our code quality, like the Do notation in the latest release.
> 
> I suggest changing this new pull request to target the branch migrate/dart3-and-fpdart (#452) instead, so we can explore the features of fpdart and apply any code improvements before we merge to develop

@marfavi i don't agree with this strategy. The app seems to run fine with a drop-in replacement, so i see no need to wait  until we have explored all its features. "Exploring" the features could take an indefinite amount of time depending on how much time we feel like investing that could be spent elsewhere. Exploring FpDart can also be done while we continue normal operations. Changing all the files where we used Dartz will also take a lot of time which can similarly be done while we are improving existing features or adding new ones. @TTA777 do you have an opinion?

marfavi (approved):
OK. I will explore code improvements on another branch

PR #471, opened 2023-05-23
fix(tickets): Fix multiple `CoffeeCard`s sharing the same product id

### LINKED ISSUE #466: 'Hero Widget' Exception is thrown when using ticket
When using a ticket, the Hero widget causes an exception to be thrown:

══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞═════════════════════════════════════════════════════════
The following assertion was thrown during a scheduler callback:
There are multiple heroes that share the same tag within a subtree.
Within each subtree for which heroes are to be animated (i.e. a PageRoute subtree), each Hero must
have a unique non-null tag.
In this case, multiple heroes had the following tag: 1
├# Here is the subtree for one of the offending heroes: Hero

marfavi:
`CoffeeCard`s are now grouped by id instead of name; if a group has tickets with different names, their names will be joined

Closes #466

### DIFF
--- a/test/features/ticket/data/datasources/ticket_remote_data_source_test.dart
+++ b/test/features/ticket/data/datasources/ticket_remote_data_source_test.dart
@@ -29,7 +29,6 @@ void main() {
 
   group('getUserTickets', () {
     test(
-      'Test that getUserTickets '
       'returns a [Right] value '
       'when the executor returns a [Right] value',
       () async {
@@ -46,7 +45,6 @@ void main() {
     );
 
     test(
-      'Test that getUserTickets '
       'returns a [Left] value '
       'when the executor returns a [Left] value',
       () async {
@@ -63,7 +61,6 @@ void main() {
     );
 
     test(
-      'Test that getUserTickets '
       'returns a [TicketCountModel] with the count of tickets and joined ticket names '
       'when the executor returns two [TicketResponse] with the same product id '
       'but different product names',
@@ -106,7 +103,6 @@ void main() {
     'useTicket',
     () {
       test(
-        'Test that useTicket '
         'returns a [Right] value '
         'when the executor returns a [Right] value',
         () async {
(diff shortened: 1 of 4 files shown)


fremartini (changes requested):
NO tests

test/features/ticket/data/datasources/ticket_remote_data_source_test.dart
```diff
@@ -28,62 +28,119 @@ void main() {
   });
 
   group('getUserTickets', () {
-    test('should return [Right] when executor returns [Right]', () async {
-      // arrange
-      when(executor.call<List<TicketResponse>>(any))
-          .thenAnswer((_) async => const Right([]));
+    test(
+      'returns a [Right] value '
```
fremartini, line 32: i still propose we follow a single format:
"should X if/when Y Z's"

fremartini (changes requested):

test/features/ticket/data/datasources/ticket_remote_data_source_test.dart
marfavi: We currently don't follow a single format:
```dart
// test/utils/time_since_test.dart
test('time since given midday and 1m30s prior returns a minute ago'), () { ...

// test/data/api/interceptors/authentication_interceptor_test.dart
test('GIVEN a token in SecureStorage WHEN calling onRequest THEN Authorization Header is added to the request', () { ...

// test/data/repositories/v2/account_repository/account_repository_test.dart
test('should call executor', () async { ...

// test/cubits/authentication/authentication_cubit_test.dart
blocTest<AuthenticationCubit, AuthenticationState>(
  'appStarted emits authenticated when a user is stored', ...
```

Although we did try to establish one in `test/README.md`:
```md
...
## Conventions

- A test description must conform to the _GIVEN_, _WHEN_, _THEN_ syntax. Example: _GIVEN no token in
  SecureStorage WHEN calling onRequest THEN no Authorization Header is added to the request_.
...
```

I've updated the names to reflect the syntax written in the README. Any further discussion about the format is out of scope for this PR

fremartini (approved):

PR #476, opened 2023-05-24
chore: bump flutter required version to 3.10.2

marfavi:


fremartini:
@marfavi a new Flutter version is available. 3.10.2, do you want to bump this instead?

fremartini (approved):

PR #478, opened 2023-05-24
fix(receipts): Remove purchase receipt view & refactor payment status display

### LINKED ISSUE #477: User shouldn't be able to view a purchase receipt when it looks so similar to a swipe receipt

marfavi:
Closes #477.
- `PurchaseReceiptListEntry` widget now has 'tappable' set to false, disabling purchase receipts.
- Timestamp has been added below the purchase/swipe status in `ReceiptListEntry`.
- Updated `PaymentStatus` string values to provide clearer descriptions.
- Renamed 'purchaseStatus' field to 'status' across multiple generic receipt widgets.
- Refactored `PurchaseReceiptListEntry` to show price differently (or hide it), based on status.
- misc: Updated `_formatter` to `_formatDate`/`_formatDateTime` in `ReceiptListEntry` and `ReceiptCard`.
- misc: Removed rounded edges from `ReceiptListEntry`.

### DIFF
--- a/lib/features/purchase/domain/entities/payment_status.dart
+++ b/lib/features/purchase/domain/entities/payment_status.dart
@@ -32,12 +32,12 @@ enum PaymentStatus {
   @override
   String toString() {
     return switch (this) {
-      PaymentStatus.completed => 'Completed',
-      PaymentStatus.rejectedPayment => 'Rejected',
-      PaymentStatus.awaitingPayment => 'Pending',
-      PaymentStatus.refunded => 'Refunded',
-      PaymentStatus.error => 'Error',
-      PaymentStatus.reserved => 'Reserved',
+      PaymentStatus.completed => 'Purchased',
+      PaymentStatus.refunded => 'Purchase refunded',
+      PaymentStatus.rejectedPayment => 'Payment rejected',
+      PaymentStatus.awaitingPayment => 'Payment pending',
+      PaymentStatus.error => 'Payment error',
+      PaymentStatus.reserved => 'Payment reserved',
     };
   }
 }
--- a/lib/features/purchase/domain/entities/payment_status.dart
+++ b/lib/features/purchase/domain/entities/payment_status.dart
@@ -1,3 +1,4 @@
+import 'package:coffeecard/base/strings.dart';
 import 'package:coffeecard/generated/api/coffeecard_api_v2.enums.swagger.dart';
 
 enum PaymentStatus {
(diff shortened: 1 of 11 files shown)


lib/features/purchase/domain/entities/payment_status.dart
```diff
@@ -32,12 +32,12 @@ enum PaymentStatus {
   @override
   String toString() {
     return switch (this) {
-      PaymentStatus.completed => 'Completed',
-      PaymentStatus.rejectedPayment => 'Rejected',
-      PaymentStatus.awaitingPayment => 'Pending',
-      PaymentStatus.refunded => 'Refunded',
-      PaymentStatus.error => 'Error',
-      PaymentStatus.reserved => 'Reserved',
+      PaymentStatus.completed => 'Purchased',
```
fremartini, line 35: perhaps these should be in Strings.dart

lib/features/receipt/presentation/widgets/list_entry/purchase_receipt_list_entry.dart
```diff
@@ -10,18 +11,32 @@ class PurchaseReceiptListEntry extends StatelessWidget {
     required this.receipt,
   });
 
+  String get priceText {
+    final price = receipt.price;
+    return switch (receipt.paymentStatus) {
+      PaymentStatus.completed => '$price,-',
```
fremartini, line 17: perhaps we should include a function formatPrice in strings.dart that handles formatting currency related strings?

lib/features/receipt/presentation/widgets/list_entry/purchase_receipt_list_entry.dart
```diff
@@ -10,18 +11,32 @@ class PurchaseReceiptListEntry extends StatelessWidget {
...
+        '($price,-)',
+      PaymentStatus.error || PaymentStatus.rejectedPayment => '',
+    };
+  }
+
   @override
   Widget build(BuildContext context) {
     return ReceiptListEntry(
-      tappable: true,
+      tappable: false,
```
fremartini, line 29: how come we no longer want to be able to open a purchase? this removes users ability to view the exact timestamp of the purchase that is displayed in the opened receipt

fremartini (changes requested):
LGTM but we need to clarify why Purchases no longer need to be able to be opened

lib/features/receipt/presentation/widgets/list_entry/purchase_receipt_list_entry.dart
marfavi: #477. The "receipt" of a purchase (especially incomplete purchases) should never resemble like a swipe receipt.

At some point we should add the purchase receipt page as specified in the v4 design spec (screenshot below)
image

fremartini: i agree, but as it stands it now removes users ability to view the exact timestamp of the purchase. Might i suggest we disable this feature when migrating to the new design?

marfavi: Then we'll just add the timestamp to the receipt list entry for now. See latest commit

lib/features/purchase/domain/entities/payment_status.dart
marfavi: Might agree, but out of scope of this PR

lib/features/receipt/presentation/widgets/list_entry/purchase_receipt_list_entry.dart
marfavi: Might agree, but out of scope of this PR

fremartini: there is already a price function in Strings.dart l. 159. I would prefer if this was fixed now while we are fixing stuff here

lib/features/purchase/domain/entities/payment_status.dart
fremartini: I would prefer if this was fixed now while we are fixing stuff here

fremartini (changes requested):
Please address my previous comments regarding Strings.dart

fremartini (approved):

PR #479, opened 2023-05-26
fix: Remove null type cast in `receipt_remote_data_source.dart`

### LINKED ISSUE #426: Avoid type casting 204 return types from API
PR #424 addressed an issue where the backend returns a Null type when the call is 204 (No Content). Our generated client doesn't take this into account (it assumes a successful request always returns the specified type). Therefore, the solution by the mentioned PR casts the return type into a nullable version of the expected return type.

A lot of our endpoints can potentially return 204, which means that our generated client could still potentially fail in cases not covered by #424. Therefore, we should fix the relevant issue https://github.com/AnalogIO/analog-core/issues/148, then revert the changes made by the PR #424.

marfavi:
Closes #426

### DIFF
--- a/lib/features/receipt/data/datasources/receipt_remote_data_source.dart
+++ b/lib/features/receipt/data/datasources/receipt_remote_data_source.dart
@@ -27,17 +27,9 @@ class ReceiptRemoteDataSource {
 
   /// Retrieves all of the users purchase receipts
   Future<Either<Failure, List<Receipt>>> getUserPurchasesReceipts() async {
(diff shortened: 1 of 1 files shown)


fremartini:
has this been fixed in the backend?

marfavi:
> has this been fixed in the backend?

Yes, https://github.com/AnalogIO/coffeecard_app/issues/426#issuecomment-1546760135

fremartini (approved):

PR #481, opened 2023-05-27
Restructure settings

fremartini:
* Move Settings to a feature
* Move nested widgets into own files


lib/features/settings/presentation/widgets/footer.dart
```diff
@@ -0,0 +1,48 @@
+import 'package:coffeecard/base/strings.dart';
```
marfavi, line 1: I'd argue this is a section alongside the profile, account, and about sections. These four widgets are the immediate widgets used in settings_page.dart and should therefore belong in widgets/sections.

marfavi (commented):
`list_entry.dart` does not belong in the Settings feature. It's a generic widget that is also used by the Receipts, Occupation, Credits, and Leaderboard features.

marfavi (approved):

PR #429, opened 2023-04-23
Improve & test reactivation authenticator

marfavi:
Closes #416

Work in progress

PR #491, opened 2023-05-30
reduce required coverage to 70%

fremartini:


marfavi:
Just merge #429 & #493 instead :)

PR #504, opened 2023-09-05
replace Redeem Voucher  with v2 endpoint

### LINKED ISSUE #500: Update voucher_remote_data_source to make use of the V2 endpoint instead
Update the implementation to make use of V2 endpoint instead. This will allow us to deprecate the old endpoint on the backend

fremartini:
closes #500


lib/features/voucher/data/datasources/voucher_remote_data_source.dart
```diff
@@ -3,23 +3,23 @@ import 'package:coffeecard/core/extensions/either_extensions.dart';
 import 'package:coffeecard/core/network/network_request_executor.dart';
 import 'package:coffeecard/features/voucher/data/models/redeemed_voucher_model.dart';
 import 'package:coffeecard/features/voucher/domain/entities/redeemed_voucher.dart';
-import 'package:coffeecard/generated/api/coffeecard_api.swagger.dart';
+import 'package:coffeecard/generated/api/coffeecard_api_v2.swagger.dart';
 import 'package:fpdart/fpdart.dart';
 
 class VoucherRemoteDataSource {
-  final CoffeecardApi apiV1;
+  final CoffeecardApiV2 apiV2;
```
marfavi, line 10: just call the field name "api". we use the distinctions in the name when we use both versions in the same class

marfavi (changes requested):
Update branch + another very small comment

lib/features/voucher/data/datasources/voucher_remote_data_source.dart
```diff
@@ -2,24 +2,24 @@ import 'package:coffeecard/core/errors/failures.dart';
...
   VoucherRemoteDataSource({
-    required this.apiV1,
+    required this.api,
     required this.executor,
   });
 
   Future<Either<NetworkFailure, RedeemedVoucher>> redeemVoucher(
     String voucher,
-  ) {
+  ) async {
```
marfavi, line 19: Async not needed here

marfavi (changes requested):
Very small

marfavi (approved):
OK

PR #512, opened 2023-09-28
Add loading indicator to splash page after duration

### LINKED ISSUE #419: Show loading spinner in the splash screen after a set delay
The splash screen is shown while loading environment and authentication information. Sometimes this takes a while (multiple seconds). Currently, the splash screen is simply a centred logo on a dark background. Adding a spinner after a fixed delay would inform the user that the app is not "stuck", and gives the illusion of the app loading faster than it does.

fremartini:
closes #419

### DIFF
--- a/lib/core/widgets/pages/splash/splash_loading_page.dart
+++ b/lib/core/widgets/pages/splash/splash_loading_page.dart
@@ -1,15 +1,54 @@
+import 'dart:async';
+
 import 'package:coffeecard/core/styles/app_colors.dart';
 import 'package:coffeecard/core/widgets/images/analog_logo.dart';
 import 'package:flutter/material.dart';
+import 'package:gap/gap.dart';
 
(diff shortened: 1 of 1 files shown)


lib/core/widgets/pages/splash/splash_loading_page.dart
```diff
@@ -1,15 +1,54 @@
...
-class SplashLoadingPage extends StatelessWidget {
+class SplashLoadingPage extends StatefulWidget {
   const SplashLoadingPage();
 
+  @override
+  State<SplashLoadingPage> createState() => _SplashLoadingPageState();
+}
+
+class _SplashLoadingPageState extends State<SplashLoadingPage> {
+  Timer? countdownTimer;
```
marfavi: use late keyword instead of nullable

marfavi (changes requested):
rest of review in person

marfavi (approved):
OK

PR #514, opened 2023-10-03
feat: sign user out after 2 hours/Never

### LINKED ISSUE #169: User should reenter his PIN code after 2 hours of inactivity
When a user has been inactive for X time or the app is restarted, a user should reenter (confirm) his password upon opening the app. E-mail is stored, so only the PIN code should be entered.

fremartini:
closes #169


lib/core/external/date_service.dart
```diff
@@ -1,4 +1,5 @@
 class DateService {
-  int currentWeekday() => DateTime.now().weekday;
```
marfavi, line 2: Does not belong in a feature branch

lib/features/authentication/data/datasources/authentication_local_data_source.dart
```diff
@@ -0,0 +1,63 @@
+import 'dart:convert';
+
+import 'package:coffeecard/features/authentication/data/models/authenticated_user_model.dart';
+import 'package:flutter_secure_storage/flutter_secure_storage.dart';
+import 'package:logger/logger.dart';
+
+class AuthenticationLocalDataSource {
```
marfavi, line 7: Don't refactor and add new features at the same time

lib/features/authentication/data/intercepters/authentication_interceptor.dart
```diff
@@ -1,25 +1,25 @@
...
+  final AuthenticationLocalDataSource localDataSource;
 
-  AuthenticationInterceptor(this._storage);
+  AuthenticationInterceptor(this.localDataSource);
 
   /// Try retrieve authentication token from storage and add authentication header if exists
   @override
   FutureOr<Request> onRequest(Request request) async {
-    final token = await _storage.readToken();
+    final token = (await localDataSource.getAuthenticatedUser())?.token;
```
marfavi, line 14: Line is too complicated.

lib/features/authentication/domain/entities/authenticated_user.dart
```diff
@@ -3,12 +3,19 @@ import 'package:equatable/equatable.dart';
 class AuthenticatedUser extends Equatable {
   final String email;
   final String token;
+  final String encodedPasscode;
+  final DateTime? lastLogin;
```
marfavi, line 7: Why would this value be nullable?

marfavi (changes requested):
The feature can be implemented with 5-10 changed files instead of 30. Remove any code refactoring from your feature branch before proceeding.

lib/features/authentication/domain/entities/authenticated_user.dart
fremartini: Because the user may not have been logged in before. Users can also sign out

lib/features/authentication/data/datasources/authentication_local_data_source.dart
fremartini: I disagree. Refactoring is part of implementing new features: leave the code prettier than when you found it.

lib/core/external/date_service.dart
marfavi: @fremartini I agree with leaving the codebase better than you found it, but consider the point of view of a reviewer. This branch's main purpose, according to the title, is to bring new functionality to the app. The reviewer would ideally only review the lines of code that directly drive this functionality.

Adding refactoring changes into the mix muddies the branch's purpose and makes the PR harder to review.  For instance, when you move files around, it is difficult to see if you only *really* changed one or two lines in the file, or if a new bug was introduced somewhere else in the code. This is because all the lines show as "added" lines in the diff.

If you believe refactoring is necessary (I don't necessarily disagree with the refactoring made in this branch), it's best to separate those changes into a dedicated branch pointing towards this one. Then, when the feature PR is merged, the refactoring PR can be merged afterwards.

This will simplify the review process by allowing reviewers to look at only one aspect of code changes at a time. Maybe someone has strong opinion about code style, but not the feature itself, or vice versa.

fremartini: @marfavi I understand it complicates the process of reviewing the code. I do personally not think this is as big an issue as you point out as booting up an editor and examining the diffs can be done rather quickly. Knowing the rate of which pull requests gets reviewed in this repository i do not believe the extra overhead of additional pull requests is worth practicing. 

On another note: i acknowledge my pull requests have lately included much refactoring which i know is unpleasant to the reviewer. I believe the restructure-the-app kinda pull requests that remain are near zero as i personally am content with the current structure we have achieved. This restructuring have in my opinion been a consequence of the entire team having to learn Flutter from scratch and how to best structure the app resulting in many try-this-architecture-out-and-see-how-it-works adventures, considering out limited experience designing software systems and apps.

fremartini:
https://github.com/AnalogIO/coffeecard_app/pull/552

marfavi:
I have a few thoughts:

## `AuthenticationCubit` vs server-side
I see a few issues with tracking session timeouts locally within `AuthenticationCubit`:

1. Right now, the session timeout depends on the device's time (`dateService.currentDateTime`). This could get tricky if someone's device time is off. I'm thinking a server-side solution might be more foolproof, as it keeps the time check consistent and out of the user's hands (and our code).

2. The code checks for session expiration only when the app starts up (`appStarted` method). What if the app stays open past the session timeout? It might be better if the app could actively monitor the session status and log out automatically when time’s up, even if it's still running.

3. Adding session timeout management into `AuthenticationCubit` increases code complexity by quite a bit. Handling this on the server-side might simplify our code and let `AuthenticationCubit` focus just whether the user is logged in or not.

I think tackling session timeouts on the server could really streamline things. It might also make features like "auto-logout after X inactive hours" easier to manage since the server would just reset the session timer with each valid request.

## Other code considerations and UX
Also, about the changes to `Dropdown` and `SettingListItem` - they seem to be doing things a bit differently now. We should probably stick to keeping these components simple to avoid any unexpected behaviors. For instance, we don't want any complex behaviour in a setting list item - see how we open a new page for other settings that requires text input or a selection from a list. How the app handles session timeouts UX-wise also needs to be discussed before we try for a new implementation. For instance, does the user even need to be able to choose the expiration time themselves?

## Next steps
I appreciate the effort you've put into this PR, but for now, I'm closing this PR. To get this feature up and running we should discuss the implementation details further in https://github.com/AnalogIO/analog-core/issues/238 before attempting to implement this client-side.

lib/core/widgets/components/dropdown.dart
```diff
@@ -2,17 +2,21 @@ import 'package:coffeecard/core/styles/app_colors.dart';
 import 'package:coffeecard/core/styles/app_text_styles.dart';
 import 'package:flutter/material.dart';
 
-class Dropdown<FilterCategory> extends StatelessWidget {
+class Dropdown<T> extends StatelessWidget {
```
marfavi, line 5: Don't change the intended use case for this widget.

marfavi (changes requested):
Address my [previous comment](https://github.com/AnalogIO/coffeecard_app/pull/514#issuecomment-1848432007).

PR #515, opened 2023-10-12
Appbundle

### LINKED ISSUE #330: Build app as appbundle
Google considers it to be best practice to build an app as appbundle and not APK.

Should be tested in Firebase.

fremartini:
Closes #330


marfavi (changes requested):
Fix merge conflicts

jonasanker:
@fremartini There has been a larger refactoring in #517 which you should take into account. It will probably give you a few bumps :)

Secondly. I would be mindful of not over-engineering with action files. I can see the value where several workflows have a need to call an action. In cases where an action is only called by one function, it might be just fine to have the steps in the workflow file :)

Thirdly, I am not sure but we should verify that Github Action supports parsing secrets as inputs such that we not leak secret values by accident when parsed as a input.

jonasanker (changes requested):
We should transition to building app bundles instead of apk's. We might need to update Firebase to support app bundle distribution with [Firebase App distribution](https://firebase.google.com/codelabs/appdistribution-app-bundles#0)

.github/actions/build_android/action.yml
```diff
@@ -0,0 +1,42 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Flutter environment
+      uses: ./.github/actions/setup_flutter_environment
+      with:
+        prod_release: ${{ inputs.prod_release }}
+
+    - name: Build apk (dev)
```
marfavi, line 25: Update name to reflect an app bundle being built

marfavi (changes requested):

jonasanker:
@fremartini Can you resolve the merge conflicts and then I'll provide my review :)

marfavi (commented):
I will leave this up to @jonasanker to review

.github/workflows/build.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and test
+name: Build application
```
jonasanker, line 1: I would prefer that the build step also invoke the test steps because they are so closely related

.github/workflows/release-dev.yml
```diff
@@ -5,17 +5,17 @@ on:
...
-      prodRelease: false
+      store_artifacts: true
+      prod_release: false
     secrets: inherit
 
-  dev_upload_ios:
+  upload_ios:
     name: Upload iOS build to Firebase App Distribution
     runs-on: ubuntu-latest
-    needs: [build_and_test]
```
jonasanker, line 18: Should we release the app if tests are failing? There might be a slight argument for Dev but I would not accept releasing the app on Prod if tests were failing

.github/workflows/release-prod.yml
```diff
@@ -2,21 +2,21 @@ name: Release Production App
...
+      store_artifacts: true
+      prod_release: true
+      tag_name: ${{ github.event.release.tag_name }}
     secrets: inherit
 
-  prod_release_ios:
+  release_ios:
     name: Upload iOS build to App Store connect
     runs-on: macos-latest
-    needs: [build_and_test]
```
jonasanker, line 19: We should not release the Production App if tests are failing, thus we need a dependency on tests

.github/workflows/test.yml
```diff
@@ -0,0 +1,31 @@
+name: Test
```
jonasanker, line 1: Possibly we will include tests as part of the build step (my argument). If we end up different, we must also run the tests pipeline on the main branch for coverage reports to function

.github/actions/build_ios/action.yml
```diff
@@ -0,0 +1,64 @@
+name: Build iOS app
+description: Build and upload iOS app
+inputs:
```
jonasanker, line 3: We should verify that secret values are not exposed when they are parsed as inputs

.github/actions/build_ios/action.yml
```diff
@@ -0,0 +1,64 @@
...
+  version:
+    required: true
+    description: "App version"
+  build_version:
+      required: true
+      description: "Build version"
+  apple_ios_signing_cert:
+      required: true
+      description: "iOS signing certificate"
```
jonasanker, line 12: Add .. in base64?

jonasanker (changes requested):
Cool refactoring. I have added some comments

.github/actions/build_ios/action.yml
fremartini: Github Actions redacts secrets automatically regardless of how they are passed around. From the logs:

2023-10-29T10:31:45.9020320Z ##[group]Run .github/scripts/setup_certs.sh
2023-10-29T10:31:45.9021020Z [36;1m.github/scripts/setup_certs.sh[0m
2023-10-29T10:31:45.9156510Z shell: /bin/bash --noprofile --norc -e -o pipefail {0}
2023-10-29T10:31:45.9157150Z env:
2023-10-29T10:31:45.9157490Z   PROD_RELEASE: false Environment variable
2023-10-29T10:31:45.9157890Z   TAG_NAME: dev <-- Environment variable
2023-10-29T10:31:45.9204640Z   APPLE_IOS_SIGNING_CERT: *** <-- Secret

2023-10-29T10:31:45.9205180Z   APPLE_IOS_SIGNING_CERT_PW: *** <-- Secret
2023-10-29T10:31:46.8769810Z   APPLE_IOS_PROVISIONING_PROFILE: *** <-- Secret

2023-10-29T10:31:46.8770530Z   APPLE_KEYCHAIN_PW: *** <-- Secret

.github/workflows/release-dev.yml
fremartini: No the app should not be released if tests are failing. Pull requests should not (and cannot?) be merged into main if tests fail. We will therefore never have a build failing the test suite on main

.github/workflows/release-prod.yml
fremartini: For the same reason above i do not believe this will be relevant

.github/workflows/build.yml
fremartini: I would prefer if the 'build' actions sole responsibility is to build the app and have a separate job for testing the app. We will not be able to merge pull requests into main if any of the pipelines fail either way. I believe this improves the readability of the actions and makes future refactoring easier.

.github/actions/build_ios/action.yml
jonasanker: <3

jonasanker (changes requested):
Tests must be run on the main branch to be able to make a code coverage diff for other branches.

.github/workflows/build.yml
jonasanker: For me, it's an edge case that Flutter supports running tests without an explicit build. 

It would have been the ideal case that tests were run on one of the build artefacts we create. Now the risk is that an app could possibly be built differently than the variant where tests are run, thus this could lead to functionality not working on a particular build.

Anyhow the argument of separation is fine. Would it then make sense to have a wrapper workflow which calls both build and test?

.github/workflows/release-dev.yml
jonasanker: My concern is that this principle is not guarded by pipelines but by a particular configuration we have for our branch protection. I think the contract is more clear when the dependency is expressed in the pipeline

.github/workflows/build.yml
fremartini: I'm not sure i see how the test build could differ from the actual build? The code is checked out and generated using the 'setup flutter environment' action for both builds. Anyhow i have made the test workflow run on pushes to the main branch

jonasanker: We are building a specific Android and iOS variant. We don't have a test case for it yet but we already have functionality which are native to the build, thus a test case could potentially behave differently on iOS vs Android.

.github/workflows/release-dev.yml
fremartini: I understand your concern but i don't see why you would be able to complete a PR with a failing pipeline to start with? That surely indicates something is wrong

jonasanker: The tests are run on the checked out branch. This might be out of sync with the master branch and thus you can't be sure that it will have the same effect on the main branch unless you enforce the constraint that branch is always in sync with main 🙂

.github/workflows/build.yml
jonasanker: The MobilePay integration reacting to the deep link. That's what my memory tells me. Really hope I'm remembering correctly 🙃

jonasanker (approved):
Oh the joy of principles. Tak Martini ✅✅✅

PR #517, opened 2023-10-12
Release dev/prod versions off a main branch

marfavi:
After merging this branch, the `develop` should be renamed into `main`, and the `production` branch should be removed.

A version is released by creating a release with a new appropriately-named tag.

Development versions would be released off of every new push onto the main branch.

**Note**: in `release-dev`/`release-prod`, `token` is deprecated. Use `serviceCredentialsFileContent` instead


.github/workflows/release-dev.yml
```diff
@@ -0,0 +1,52 @@
+name: Publish Development App
+
+on:
+  release:
+    types: [created]
```
TTA777, line 5: Should this not be on all merges to the branch instead?

.github/workflows/release.yml
```diff
@@ -1,123 +0,0 @@
...
-      - name: Upload to Google Play Store
-        uses: r0adkll/upload-google-play@v1.0.16
-        with:
-          serviceAccountJsonPlainText: ${{ secrets.PLAYSTORE_SERVICE_ACCOUNT_JSON }}
-          packageName: dk.analog.digitalclipcard
-          releaseFiles: ${{ steps.sign_app.outputs.signedReleaseFile }}
-          track: internal
-          status: draft
-
-  tag:
```
TTA777, line 111: Are we no longer making use of semantic versioning tags for releases?

TTA777 (changes requested):

.github/workflows/release-dev.yml
jonasanker, line 5: The trigger here should be a commit on the main branch for dev releases. Only GitHub releases should trigger a prod deployment

.github/workflows/release-prod.yml
```diff
@@ -0,0 +1,59 @@
...
+    uses: ./.github/workflows/build.yml
+    with:
+      storeArtifacts: true
+    secrets: inherit
+
+  prod_release_ios:
+    name: Upload iOS build to App Store connect
+    runs-on: macos-latest
+    needs: [build_and_test]
+    if: github.ref_name == 'production'
```
jonasanker, line 18: This condition will not be made when we are triggered by a new release

jonasanker (changes requested):
To make this work also the build flow must be changed. Previously we signed the apps for prod builds, and handled it differently for the dev app. Since the condition for production branches is not valid any more, we must rewrite the build part again to handle two types of building the app

.github/workflows/release.yml
jonasanker: This would not be needed as a GitHub release is also a git tag, thus the value of the tag will be determined upon creation of the release not in the pipeline

.github/workflows/release-dev.yml
marfavi: Thanks for catching this

marfavi: Thanks for catching this

.github/workflows/release-prod.yml
marfavi: Thanks for catching this

jonasanker (dismissed):
LGTM 😇

jonasanker (approved):

PR #522, opened 2023-10-15
Update Android SDKs, Java JDK and action versions to newer versions

jonasanker:


.github/workflows/build.yml
```diff
@@ -167,7 +167,7 @@ jobs:
         uses: actions/checkout@v2.4.0
 
       - name: Setup Flutter environment
-        uses: subosito/flutter-action@v1.5.3
```
fremartini, line 170: i think we should pin the version to avoid stuff breaking because of new versions.

fremartini (changes requested):
i think we should pin the version to avoid stuff breaking because of new versions.

.github/workflows/build.yml
jonasanker: I don't disagree as such. However we also be fine to lock into just the major or minor version as you wouldn't normally expect breaking behaviour between patches .

jonasanker:
> i think we should pin the version to avoid stuff breaking because of new versions.

I don't disagree as such. However we also be fine to lock into just the major or minor version as you wouldn't normally expect breaking behaviour between patches

fremartini:
> > i think we should pin the version to avoid stuff breaking because of new versions.
> 
> I don't disagree as such. However we also be fine to lock into just the major or minor version as you wouldn't normally expect breaking behaviour between patches.
> 
> Could we maybe just lock in in the minor version?

Minor versions would be fine. Expectations and reality have a tendency to not align

jonasanker:
> > > i think we should pin the version to avoid stuff breaking because of new versions.
> > 
> > 
> > I don't disagree as such. However we also be fine to lock into just the major or minor version as you wouldn't normally expect breaking behaviour between patches.
> > Could we maybe just lock in in the minor version?
> 
> Minor versions would be fine. Expectations and reality have a tendency to not align

It seems you can not refer to a minor version. Only major versions or absolute versions are supported, thus I have changed all actions to refer their latest version.

@fremartini Please re-review

fremartini (approved):

PR #413, opened 2023-03-28
feat: barista perks

fredpetersen:
Closes #422 

Visual: A new section has been added to the tickets page, that shows your perks if your user is registered as Barista+

All purchaseable products are now loaded in the SplashRouter which has now been renamed to MainRedirectionRouter (These products are now also fetched from ProductsControllerV2 instead of V1)

PR #528, opened 2023-10-24
refactor: load user info before main content flow

marfavi:
This commit ensures that the user is loaded from the "splash router" before the home page is shown.
This means everything below the "home page" widget will have access to a loaded user cubit state.

Extracted from #413 and therefore also a prerequesite for that PR.


marfavi (commented):
Should ensure test coverage in changed lines before proceeding

fredpetersen (approved):
👍

PR #529, opened 2023-10-30
Fix Android appbundle path

fremartini:
Building the Appbundle puts the artifact in a different folder compared to the APK


jonasanker (approved):
Lgtm

PR #530, opened 2023-10-30
Fix Android appbundle artifact name

fremartini:
Note: the [prod version](https://github.com/AnalogIO/coffeecard_app/blob/main/.github/workflows/release-prod.yml) does not look like it needs any update. If I am mistaken, please correct this PR


jonasanker (approved):
Lgtm vol 2 ✅✅

PR #541, opened 2023-11-14
refactor: opening hours

fremartini:
Fix opening hours to match the cafe

Uses the flutter-native TimeOfDay class, which has localized time formatting.
This has been extended with comparison methods <= and isInTimeslot.
Timeslot class simplifies with this change.

Simplify DateService to only contain one getter.

Make sure of fpdart's Option instead of null values. Use None for closed days.

Clean up tests, add more tests, and use NiceMocks for mocking.


lib/core/external/date_service.dart
```diff
@@ -1,4 +1,6 @@
 class DateService {
-  int currentWeekday() => DateTime.now().weekday;
-  int currentHour() => DateTime.now().hour;
+  DateTime now() => DateTime.now();
```
marfavi, line 2: Minor gripe but I'd suggest to use getters instead

lib/features/opening_hours/domain/entities/timeslot.dart
```diff
@@ -2,17 +2,31 @@ import 'package:coffeecard/core/strings.dart';
 import 'package:equatable/equatable.dart';
 
 class Timeslot extends Equatable {
-  final int? start;
-  final int? end;
+  final (int, int)? start;
```
marfavi, line 5: Consider creating a typedef with named fields ("hours", "minutes") so it's more clear that this tuple represents a time of day

lib/features/opening_hours/domain/entities/timeslot.dart
```diff
@@ -2,17 +2,31 @@ import 'package:coffeecard/core/strings.dart';
...
 
   const Timeslot({this.start, this.end});
 
   bool get isClosed => start == null || end == null;
 
   @override
-  String toString() => isClosed
-      ? Strings.closed
-      : '${start.toString().padLeft(2, '0')}:00 - $end:00';
+  String toString() {
```
marfavi, line 13: Remove in favour of having no timeslot for closed days

marfavi (changes requested):

PR #542, opened 2023-11-16
fix(build): Android duplicate signing

fremartini:
https://stackoverflow.com/questions/5089042/jarsigner-unable-to-sign-jar-java-util-zip-zipexception-invalid-entry-compres

"I  encountered this when signing my .aab file. Removing the duplicate signing (once as part of the bundling, once manually) fixed it. This was part of the default react-native app scaffolding.

The app/build.gradle file includes a section android/buildTypes/release which had its signingConfig key set. When generating .apk files it seemed to be ignored but when switching to .aab format it looks like it did apply that signing. When I then did my own signing in CI, it complained because it was already signed."


marfavi (approved):
LGTM

PR #557, opened 2023-11-30
fix: various things

marfavi:
- Fix whitespace in section title rows that have a widget next to them by adding a new constructor.
- Fix scrim text when claiming a free product or purchasing a single product.
- Fix "swipe to confirm" text colour.
- Fix barista perks section being present when no perks are available.
- Fix "Purchase tickets" page exiting when a purchase is cancelled or rejected.
- Fix single purchases not auto-claiming the ticket.
- Fix single purchases not showing a swipe receipt.
- Fix miscellaneous whitespace issues.
- Fix heading text line height being too high.
- Fix SectionTitle not being used in  spots it should.
- Fix Grid sometimes containing empty space at the bottom of its content.


fremartini (dismissed):

fremartini (approved):

marfavi:
> thats not really what i meant. The height is fine but i would like the 1.1 to be put into a variable called lineHeight
> 
> const lineHeight = 1.1; height = ....

Not sure how this improves things. The new `const lineHeight = ...` line works like the line above it to minimize conditional logic in the return statement. If you mean something else, can you show a complete example?

fremartini (approved):

PR #558, opened 2023-12-03
run tests in Docker container / generate goldens

### LINKED ISSUE #382: Docker file for golden generation
Currently goldens only pass `flutter test` step in pipeline, if generated by a linux machine.

There should be a docker file readily available for generating golden files, so we're not dependant on our linux developers when changing our ui

fremartini:
closes #382


marfavi:
#587 removes goldens

PR #560, opened 2023-12-08
fix(ci): development version built in release pipeline

marfavi:
This PR fixes the release pipeline bug that caused building a development version for release (in the first commit), and increases code robustness to prevent the same bug from appearing again (in the second commit).

The error was caused by a bug in actions code, which passed an ill-defined input variable (whether to build a dev/prod app) to another action. The input variable expected by the second action was optional, defaulting to build a dev version if no input variable is present.

---

Build configurations across Android and iOS build pipelines were updated to enforce the specification of 'development' or 'production' as mandatory build types. Introduced a shared 'verify_build_type' action that validates this input. This change ensures that artifacts are built against the correct backend environments and clarifies the pipeline behavior by removing ambiguity in build type selection. Additionally, updated steps to reflect the new build type validation and adjusted environment setup actions to accommodate the enforcement of build types.

Finally, this PR fixes the following yml syntax/schema errors:
- `inputs.<id>.options` in an action
- `env` on the root level of an action

### DIFF
--- a/.github/actions/setup_flutter_environment/action.yml
+++ b/.github/actions/setup_flutter_environment/action.yml
@@ -21,7 +21,7 @@ runs:
     - name: Setup Flutter environment
       uses: subosito/flutter-action@v2.11.0
       with:
-        flutter-version: 3.13.7
+        flutter-version: 3.16.0
         channel: "stable"
 
     - name: Download dependencies
--- a/.github/actions/setup_flutter_environment/action.yml
+++ b/.github/actions/setup_flutter_environment/action.yml
@@ -1,29 +1,35 @@
 name: Setup Flutter environment
 description: Setup the Flutter environment
 inputs:
-  prod_release:
-    required: false
-    default: 'false'
-    description: "Is prod release"
-
-env:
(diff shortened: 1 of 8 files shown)


.github/actions/setup_flutter_environment/action.yml
```diff
@@ -1,32 +1,38 @@
...
+    - name: Validate build_type
+      uses: ./.github/actions/verify_build_type
+      with:
+        build_type: ${{ inputs.build_type }}
+
     - name: Setup Flutter environment
       uses: subosito/flutter-action@v2.11.0
       with:
-        flutter-version: ${{ env.FLUTTER_VERSION }}
+        flutter-version: 3.16.0
```
marfavi, line 24: Fun fact: since using `env` is not valid at the root level of an actions file, fixing this (by moving the flutter version number to this position) caused build errors caused by using an outdated version of Flutter. So, until now, the FLUTTER_VERSION env variable was ignored by the pipeline.

I've updated the version number to match the one in the top-level README.md

fremartini (approved):

PR #575, opened 2024-11-05
build: bump packages and dart/flutter version

marfavi:
Update packages, dart/flutter versions and update files to be able to build. This includes addressing changes to chopper (http client), which had breaking changes to its interceptor. This class can no longer be easily tested, so its tests have been removed.


TTA777 (approved):
lgtm

PR #584, opened 2024-12-03
fix(build): allow non-modular includes in framework modules

marfavi:
Allows build for iOS on Macbooks


Slug-Boi (approved):
Looks good fixes the build error 👍🏻

PR #589, opened 2025-05-05
Listen to logins and remove passcode flow

### LINKED ISSUE #125: (New flow) Log in and register via Magic links
LoginFlow would be replaced by a email text field.
A magic link is always sent to that email.
- If that email belongs to an account, the link will log the user in.
- Otherwise, the link will trigger an account registration.

This will retire the use of passcodes.

When a user registers using a magic link, they will skip the current two screens of the registration flow, as the email is already known (and passcodes will retire). The user would just need to provide their name and programme (#124)

image

A-Guldborg:
Closes #125 

@marfavi with @TTA777 we figured out some initial steps and for now you should be able to do two things:

1. Open urls on Android phones (url: `analogcoffeecard-dev://auth?token=<UUID>`)
2. Send a magic link when entering mail in the login field on initial post. 

We need to distinguish between two intent-filters in Android and IOS to ensure we handle mobilepay and authentication deeplinks differently.

Base of PR is #588 since this branch is what works for me currently, but should be updated to main once #588 is merged.


marfavi:
Base is now #587

marfavi:
Base is now main. Please resolve merge conflicts @A-Guldborg

PR #599, opened 2025-05-25
fix(ui, android): back button exits app; navBar tap scrolling and going back at the same time

### LINKED ISSUE #578: Native "Back" action on Android closes app
**Describe the bug**
When swiping from right to left ("Back" gesture) on Android, the app closes instead of going back, for example, when displaying a receipt. Thus, a user has to reach for the in-UI Back element instead of the native back action.

**To Reproduce**
1. Go to "Receipts"
2. Open a receipt
3. Perform "Back" gesture (Swipe from right to left)

**Expected behavior**
The app should navigate back in history.

**Device (please complete the following information):**
 - Device: Pixel 4A
 - OS: Android 14
 - App Version: 3.9

### LINKED ISSUE #600: Visual jank when pressing currently visited tab in navigation bar
**Describe the bug**
When tapping one of the four navigation bar icons, if you were to tap the icon of the flow you were already on, the following behaviour should happen:

- If you can go back in the flow (i.e. back button is visible in the top left corner), then the tap should bring the user back until the root is reached.
- If you were already at the root, then the tap should scroll the user to the top of the page

Currently, both happen at the same time, which can cause some visual jank.

**To Reproduce**
1. Tap the Receipts tab in the navigation bar to go to Receipts
2. Scroll down the page and click on an old swipe receipt
3. Tap the Receipts tab
4. Observe visual jank (app both goes back in flow and scrolls up at the same time)

**Expected behavior**
App should only go back in the flow (to the Receipts list view).

**Screenshots**
https://github.com/user-attachments/assets/16440a74-c1cd-4481-9d46-af112ff14a2b

**Device**
Android emulator

marfavi:
Fixes #578, fixes #600

### DIFF
--- a/lib/core/widgets/pages/home_page.dart
+++ b/lib/core/widgets/pages/home_page.dart
@@ -63,7 +63,7 @@ class _HomePageState extends State<HomePage> {
     // Pop the current flow index off the stack
     final _ = _navFlowsStack.removeLast();
 
-    // Exit app if the stack is now empty, or change the the page
+    // Exit app if the stack is now empty, or change the page
     if (_navFlowsStack.isEmpty) {
       SystemNavigator.pop(animated: true);
     } else {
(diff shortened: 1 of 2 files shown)


marfavi:
@duckth, the regression is now fully fixed

duckth (dismissed):

duckth (approved):

PR #607, opened 2026-04-17
fix: bump and address breaking change in DottedBorder

duckth:
Bump packages since there was a version incompatibility before.
DottedBorder has breaking changes since 3.0.0 and this also addresses that

### DIFF
--- a/.github/actions/setup_flutter_environment/action.yml
+++ b/.github/actions/setup_flutter_environment/action.yml
@@ -19,9 +19,9 @@ runs:
         build_type: ${{ inputs.build_type }}
 
     - name: Setup Flutter environment
-      uses: subosito/flutter-action@v2.11.0
+      uses: subosito/flutter-action@v2.23.0
       with:
-        flutter-version: 3.32.0
+        flutter-version: 3.41.7
         channel: "stable"
 
     - name: Download dependencies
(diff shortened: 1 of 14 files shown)


marfavi (changes requested):
Version solving fails in pipeline, please address this:
```
Because no versions of envied_generator match >1.3.4 <2.0.0 and envied_generator 1.3.4 requires SDK version ^3.9.0, envied_generator ^1.3.4 is forbidden.
So, because coffeecard depends on envied_generator ^1.3.4, version solving failed.
Failed to update packages.
```

duckth:
Aaaaaargghhhhhhh
```
   info • The type parameter name 'Body' matches a visible type name • lib/features/authentication/data/intercepters/authentication_interceptor.dart:13:38 • avoid_types_as_parameter_names
   info • 'groupValue' is deprecated and shouldn't be used. Use a RadioGroup ancestor to manage group value instead. This feature was deprecated after v3.32.0-0.0.pre • lib/features/occupation/presentation/widgets/occupation_form.dart:123:9 • deprecated_member_use
   info • 'onChanged' is deprecated and shouldn't be used. Use RadioGroup to handle value change instead. This feature was deprecated after v3.32.0-0.0.pre • lib/features/occupation/presentation/widgets/occupation_form.dart:126:9 • deprecated_member_use
```


@marfavi help

